PR #8, opened 2020-08-03
probable fix for mismatched dep versions

drewbanin:
TODO before merging:
 - test and verify this change

### DIFF
--- a/hubcap.py
+++ b/hubcap.py
@@ -247,6 +247,11 @@ for org_name, repos in TRACKED_REPOS.items():
 
                 version_path = os.path.join(repo_dir, "{}.json".format(tag))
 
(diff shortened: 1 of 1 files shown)


VersusFacit (commented):
Closing this out after a short chat with Drew.

One, the context from Drew:
> i think the issue was that hubcap was looking at master  to figure out child dependencies of a package and it was yielding incorrect dependencies for packages that didn’t branch off of (and merge back into) a master branch

Two, I'm pretty it's been address by how we're obtaining package version dependencies now (based on parsing the yaml). We may find we need to import some kind of tag checking out, but that logic would go _before_ the parse_packages step, whereas this is after.

PR #17, opened 2020-10-01
Add spark-utils

jtcohen6:
This package leverages "dispatched" macros (new in v0.18.0) to get other packages running on a non-core plugin (Spark). I hope this might be a model for other community-contributed plugins + packages.


drewbanin (approved):
How exciting! Ship it :)

PR #22, opened 2020-11-06
make hubcap unaware of branch names

drewbanin:
- Previously, hubcap used dbt's git client to clone repos. Because hubcap does not provide a branch name, the dbt git client would try to check out a branch called `master`
- Now, we just call `git clone` directly, so whatever the primary branch is for the repo should be cloned and checked out

Confirmed that this worked by adding a new test project (branch=`main`) to the hub site locally

### DIFF
--- a/hubcap.py
+++ b/hubcap.py
@@ -183,8 +183,10 @@ for org_name, repos in TRACKED_REPOS.items():
             git_path = os.path.join(TMP_DIR, repo)
 
             print("Cloning repo {}".format(clone_url))
-            dbt.clients.git.clone_and_checkout(clone_url, cwd=TMP_DIR, dirname=repo)
+            if os.path.exists(git_path):
+                dbt.clients.system.rmdir(git_path)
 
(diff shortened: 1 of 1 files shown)


clrcrl:
What about [these lines](https://github.com/fishtown-analytics/hubcap/blob/001c4197eacd04e6079560be1ba066b6571ee714/hubcap.py#L53-L54)? I can see `master` hardcoded there

drewbanin:
Those lines clone the `master` branch of the `hub.getdbt.com` repo itself. It's the temporary workspace where we create new branches for each package to update. At the end of the hubcap script, we push all of these new branches for updated packages (branched off of master) to github and open PRs for them.

This is definitely super confusing and I would be strongly in favor of doing this way differently in the future! Don't think the lines you highlighted impact this particular issue though

clrcrl:
Got it, thanks for clarifying!

clrcrl (approved):
LGTM

drewbanin:
@clrcrl merging this - please lmk if we you see any issues when new packages/versions are added to the hub site in the future!

PR #28, opened 2021-01-14
add tsql-utils to hub

swanderz:
would love to streamline install experience for dbt-msft end users!


jtcohen6 (approved):
Thanks @swanderz!

PR #40, opened 2021-06-16
Add re-data/dbt-re-data package

mateuszklimek:
Adding package computing models for data quality metrics & anomalies in them.

More info:
https://github.com/re-data/dbt-re-data and https://github.com/re-data/re-data

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -137,5 +137,8 @@
     ],
     "entechlog": [
         "dbt-snow-mask"
+    ],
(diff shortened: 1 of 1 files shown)


jtcohen6 (approved):
Very neat, @mateuszklimek. To be clear, I haven't looked at the package extensively, but it sounds like you're attempting a solution for a real problem. Happy to get this on the Hub site.

PR #43, opened 2021-06-30
Rename org: fishtown-analytics → dbt-labs

jtcohen6:
For backwards compatibility, I think the move should be for the Hub site to accept both:
```
packages:
  - package: fishtown-analytics/dbt_utils
  - package: dbt-labs/dbt_utils
```
GitHub can handle the redirect for the former. We can hide the Fishtown-named ones from the Hub site homepage.

@drewbanin What do you think? Am I missing something big here?


jtcohen6:
Alternatively, we could remove the `fishtown-analytics` packages entirely if we want to force folks to switch to `dbt-labs` to get new versions.

drewbanin:
> What do you think? Am I missing something big here?

My only real concern is around whether dbt's package resolver will "know" that `fishtown-analytics/dbt_utils` and `dbt-labs/dbt_utils` are the same package. I think that it will not, and the failure mode would come from recursive dependencies. Just say:
 - `fishtown-analytics/snowplow` depends on `fishtown-analytics/dbt_utils`
 - you depend on `dbt-labs/dbt_utils` directly

In this scenario, I _think_ dbt will fail to install these packages correctly. I feel 80% sure about that, so it's worth testing!

drewbanin:
I guess I should say: I don't view it as a blocker to merging this PR, but we should have a think about the migration path and how to help users avoid traps like this one if it is indeed going to be a challenge

jtcohen6:
That's a really good point. Worth looking into, and worth updating all our packages with dbt-utils dependencies to point to `dbt-labs/dbt_utils` instead.

This makes me glad we've held off on some of the v0.20-related package releases (https://github.com/dbt-labs/dbt-utils/issues/381)!

jtcohen6:
@drewbanin Here's my latest thinking:

- You're absolutely right, dbt won't be able to resolve dependencies for `fishtown-analytics/dbt_utils` vs. `dbt-labs/dbt_utils`
- We should start pushing new package versions to `dbt-labs` org _only_ (i.e. remove `fishtown-analytics` from hubcap, while leaving old versions in hub.getdbt.com)
- Our next package versions—which we need for dbt v0.20 + dbt-utils v0.7 compatibility, anyway—should explicitly depend on `dbt-labs/dbt_utils` instead of `fishtown-analytics/dbt_utils`. We call out, in those package releases, that users _must_ switch any `fishtown-analytics` packages to `dbt-labs` for dependency resolution.
- There might still be some friction with packages we don't maintain that depend on `fishtown-analytics/dbt_utils`. Ifthat comes up, we can encourage the maintainer to cut a new patch release that switches to `dbt-labs`.

Does that track?

drewbanin:
> We should start pushing new package versions to dbt-labs org only (i.e. remove fishtown-analytics from hubcap, while leaving old versions in hub.getdbt.com)

Agree!

> Our next package versions—which we need for dbt v0.20 + dbt-utils v0.7 compatibility, anyway—should explicitly depend on dbt-labs/dbt_utils instead of fishtown-analytics/dbt_utils

Also agree!

> There might still be some friction with packages we don't maintain that depend on fishtown-analytics/dbt_utils. If that comes up, we can encourage the maintainer to cut a new patch release that switches to dbt-labs.

Agree x3. We can see the list of packages on the hub site that depend on dbt-utils. Maybe we can create issues to notify the maintainers? Here's a handy one-liner you can run from the root of the hub.getdbt.com repo:
```
grep -ril fishtown-analytics/dbt_utils data/packages | awk -F '/' '{ print $3 "/" $4 }' | sort | uni
```

It is... pretty long!

```
Datavault-UK/dbtvault
Montreal-Analytics/snowflake_utils
calogica/dbt_date
fishtown-analytics/adwords
fishtown-analytics/audit_helper
fishtown-analytics/bing_ads
fishtown-analytics/codegen
fishtown-analytics/dbt_external_tables
fishtown-analytics/dbt_utils
fishtown-analytics/facebook_ads
fishtown-analytics/fishtown_analytics_ecommerce
fishtown-analytics/heap
fishtown-analytics/logging
fishtown-analytics/outbrain
fishtown-analytics/purecloud
fishtown-analytics/redshift
fishtown-analytics/segment
fishtown-analytics/snowplow
fishtown-analytics/stitch_utils
fishtown-analytics/taboola
fishtown-analytics/zendesk
fivetran/asana
fivetran/asana_source
fivetran/facebook_ads_source
fivetran/fivetran_log
fivetran/github
fivetran/github_source
fivetran/google_ads_source
fivetran/greenhouse_source
fivetran/hubspot
fivetran/hubspot_source
fivetran/intercom_source
fivetran/jira_source
fivetran/klaviyo_source
fivetran/lever_source
fivetran/linkedin
fivetran/linkedin_source
fivetran/mailchimp
fivetran/marketo
fivetran/marketo_source
fivetran/microsoft_ads_source
fivetran/mixpanel
fivetran/netsuite
fivetran/netsuite_source
fivetran/pardot_source
fivetran/pinterest_source
fivetran/quickbooks_source
fivetran/salesforce
fivetran/salesforce_formula_utils
fivetran/salesforce_source
fivetran/shopify_source
fivetran/snapchat_ads_source
fivetran/stripe_source
fivetran/twitter_ads_source
fivetran/xero_source
fivetran/zendesk_source
gitlabhq/snowflake_spend
kristeligt-dagblad/dbt_ml
re-data/re_data
sgoley/postgres_utils
tailsdotcom/dbt_artifacts
tnightengale/dbt_meta_testing
```

jtcohen6:
@drewbanin Thanks for validating the approach. I just removed the `fishtown-analytics` org from `hub.json` entirely, leaving only `dbt-labs`.

> Maybe we can create issues to notify the maintainers?

My plan is to write a discourse post about the migration, post it in the dbt Community Slack (probably #dbt-package-releases), and also share it directly with the package maintainers who need to make bigger lifts.

I'm going to merge this in to kickstart the process. If all goes well, we should get to spend this evening watching a bunch of PRs open and promptly merge themselves in the hub.getdbt.com repo.

PR #45, opened 2021-07-07
Checkout tag before making package spec

### LINKED ISSUE #44: Get package dependencies for specific release tag
(Similar to #21)

Hubcap uses the default branch, rather than the specific release, to get information about `packages` (dependencies). Instead, for each tagged release it's adding, it should check out that tag before introspecting packages.

This is causing issues right now for `fivetran_utils`. They can't change the default branch (it's pointed to by some older versions of other packages), so they want to cut releases from non-default branches instead. Everything works except for the `packages` dict created by Hubcap.

jtcohen6:
resolves #44

@drewbanin I wrote a few hopeful lines of code :) I'm less clear on how to reasonably test this.

### DIFF
--- a/hubcap.py
+++ b/hubcap.py
@@ -274,7 +274,10 @@ for org_name, repos in TRACKED_REPOS.items():
                     continue
 
                 version_path = os.path.join(repo_dir, "{}.json".format(tag))
-
+                
+                print("    Checking out tag {}".format(tag))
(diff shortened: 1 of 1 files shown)


drewbanin (approved):
LGTM to me - going to merge and watch the logs in prod. Thanks @jtcohen6!

Issue #53, opened 2021-07-29
Generate project specification without dbt


VersusFacit:
## Problem
dbt expectations cannot merge the newest version of their package to `hub.getdbt.com` because the dbt version employed by this repo's build script requires an incompatible version of core.

The script errors out at setup.

## Background

To generate project specifications (i.e. the information used to generate project pages on https://hub.getdbt.com/), `hubcap.py` uses `dbt` itself to run shell commands and extract project information. 

Among other things, this requires a phony dbt profile and a specific dbt version (see `/requirements.txt`). This introduces possible conflict with any packages using a `require-dbt-version` configuration tag in their `dbt_project.yml`. 

As mentioned, dbt expectations is blocked from pushing their newest release because their [pinned dbt version](https://github.com/calogica/dbt-expectations/blob/e47358d5ee5f162457fd4b40aa9ea84d8879ea93/dbt_project.yml#L8) is not compatible with the version of dbt we use in the hubcap build script.

![image](https://user-images.githubusercontent.com/67295367/127452318-d7e7c448-5a10-4250-b188-92d0a344edb2.png)

## Solution

All functionality performed by dbt can be done using familiar Python system libraries and yaml parsing files. Let's remove Core from this build script.

Thankfully, the two files which hold this information are both yaml files located by design at root level of a dbt project--`dbt_project.yml` and `packages.yml`. Yaml parsing libraries should do the trick to put together a specification and a dependency chain.

As a bonus, any modularization of these components will be appreciated in the now and the future.

## Outcomes

- we unblock the dbt expectations team
- we insulate ourselves from this danger

clausherther:
Hello! Just wanted to check in on when we might be able to merge this to get dbt hub back in sync? Thanks!

VersusFacit:
Addressed by #61 . dbt expectations has been back in sync after a manual merge some weeks ago, but also will now automatically sync as before.

PR #54, opened 2021-07-29
Streamline code to about the halfway point.

VersusFacit:
Per the request of the [Asana ticket](https://app.asana.com/0/1200663523005279/1200674036437001/f), this is starting off as a WIP to track progress.

I am about halfway stripping references to the dbt hackathon command library. It's used at most levels of the script so I'm taking time to make sure output looks good at each step. I won't do too much modularizing on this PR but will aim to chunk out obvious sections.


annafil:
Makes sense to me so far and thank you for documenting this so well! 👏 

Will hold detailed feedback till you mark this as ready :) You can also mark this PR as WIP for now and then mark it for review when you're ready.

VersusFacit:
In the output of a more-or-less full sweep, I noticed this script will pull down and apparently push up the 0.4.1 release of dbt epxectations--the original bug that has spurred this refactor. Hence, this refactor closes #53 

**How did you test it?** I admit, not as robustly as I'd like to in the future, but this requires further restructuring and more complex logic for disabling `git` commands by using of `-n`. Hence, my local testing was done incrementally and pseudo-holistically
- incrementally: comparing output of different functions throughout every little step development
- pseudo-holistically: at the end, I ensured code could spin through all iterations without error (or minor error as a result of obscure errors; for example, one dbt-zuora repo is empty and that particular case isn't yet handled, just skipped over)

**Why was a refactor needed?** We needed to remove _every_ reference to `dbt.*`. It was used at all levels of the code as a convenience for handling subshells. My own syscall solution is mediocre and can be improved, but it is reliable enough to get the job done (I can provide an argument here if desired).

**How did the code run.** I am confident the functionality of the code itself OTHER THAN the git functions is working as desired. I can't quite test the git functions beyond using `-n` flags and comparing outputs. Honestly, there may be strange pathing caveats, since the `run_cmd` dbt client function is unclear about how it opens subshells (and thus what CWD should be at any time).

However, given that the script has been working, we can reasonably trust me leaving those commands untouched should remain working (assuming I did not introduce bugs while migrating to use of the `subprocess` library).

##  Proposed Deployment
I'll have to deploy this onto Heroku, but I also think that for running this after code review, it should be done locally so I can track it's progress and catch bugs as its running. This is going to be a very stateful/destructive process and I'd like to limit its operation to a subset of repos or even 1 by 1 until it's clear the script is acting within spec.

VersusFacit:
Bug found as a result of https://github.com/dbt-labs/hub.getdbt.com/pull/789/files

PR #61, opened 2021-09-01
Hubcap revamp take 2

VersusFacit:
# Development Notes

**Purpose**: Primarily, remove the dependency on `dbt` itself. Secondarily, modularize and add some SWE-ing tuneups into the mix.

**Perks**:
* no more dbt dependency
* the code overall is more modular and makes use of more Python builtins; you'll see this reflected in logging
* script defaults to a development environment--i.e. it's safe to run this script locally and expect no stateful changes to our assets; ENV=test or ENV=prod is required to see actual branch pushes in the hub repo, but the default is ENV=development (this variable is set/overridden in Heroku's environment variables console)
* reduced the number of directory changes to as few as possible and reduced `continue` logic where possible to streamline execution
* added some light documentation to help the future reader through execution

Note: I tried several different combinations of commands to have git recognize the lineage between `hubcap.py` and `hubcap/hubcap.py` but git _really_ wants to treat them as distinct. Spliting the move and alteration steps across commits may preserve history in branch but a merge forgets it.

# Testing

There were several PRs already up (mostly from Fivetran), so I took the opportunity to compare both the `index.json` and the version specs this PR's code is creating vs. production's artifacts. `diff` tells me they are a one to one match :) 

Sample spec from my script is the same (save timestamps) as the one created by production.
```json
{
    "id": "fivetran/stripe_source/v0.4.2",
    "name": "stripe_source",
    "version": "v0.4.2",
    "published_at": "2021-09-01T00:11:57.250971-07:00",
    "packages": [
        {
            "package": "fivetran/fivetran_utils",
            "version": [
                ">=0.2.0",
                "<0.3.0"
            ]
        }
    ],
    "works_with": [],
    "_source": {
        "type": "github",
        "url": "https://github.com/fivetran/dbt_stripe_source/tree/v0.4.2/",
        "readme": "https://raw.githubusercontent.com/fivetran/dbt_stripe_source/v0.4.2/README.md"
    },
    "downloads": {
        "tarball": "https://codeload.github.com/fivetran/dbt_stripe_source/tar.gz/v0.4.2",
        "format": "tgz",
        "sha1": "c2f2c1422901e1943cb4cbed3511e4f63f1dec57"
    }
}
```

I can also share a full build log I captured on my local machine if wanted! Here's an excerpt
```
2021-09-01 00:19:16 INFO     preparing script state
2021-09-01 00:19:16 INFO     cloning hub site repo to /Users/mila_fa/Dev/dbt-hubcap-staging/git-tmp/hub
2021-09-01 00:19:17 INFO
2021-09-01 00:19:17 INFO     Your branch is up to date with 'origin/master'.
2021-09-01 00:19:18 INFO     Already up to date.
2021-09-01 00:19:18 INFO     preparing branches for packages with versions to be added
2021-09-01 00:19:18 INFO     cloning package dbt-utils maintained by dbt-labs to /Users/mila_fa/Dev/dbt-hubcap-staging/git-tmp/dbt-utils
```

Better yet, go survey the lovely PRs in hub :) 
* [dbt expectations 0.4.2](https://github.com/dbt-labs/hub.getdbt.com/pull/849)
* [jitsu](https://github.com/dbt-labs/hub.getdbt.com/pull/850) 

Also, the run didn't result in extra PRs being added to hub even though it did tech the PRs existed for `stripe_source`, `fivetran_utils`, and `youtube_analytics_source`. Confirmation we have achieved idempotent state changes (which is great news).

# Experience it yourself, if ya' like

**Want to see this work in action for yourself, logs and all?** Easy! Run `heroku run --app=dbt-hubcap-staging ./cron.sh` in a heroku-enabled terminal (an honestly easy install setup). 

This will call the code on the staging machine which is the newest commit on this branch. New branches will be pushed to hub.


VersusFacit:
Other things not addressed in this that should be considered in the future:
* code still has a bit of logic leftover from when we were doing CLI deployments and this could be removed at a future date
* logging could be tweaked in places to avoid extra whitespace
* more continue logic could be streamlined

hubcap/setup.py
```diff
@@ -0,0 +1,58 @@
...
+    level=logging.INFO,
+    datefmt='%Y-%m-%d %H:%M:%S'
+)
+
+PKG_EXCLUSION_OVERRIDES = {
+    'dbt-labs': ['bing-ads', 'recurly', 'shopify', 'purecloud', 'zendesk', 'stripe', 'outbrain', 'taboola',
+                 'quickbooks', 'bing-ads', 'taboola', 'heap', 'ecommerce', 'facebook-ads', 'zendesk', 'stripe',
+                 'purecloud'],
+    'beautypie': ['dbt-google-analytics-360', 'dbt-zendesk-support']
+}
```
jtcohen6, line 25: Could we make the exclude list another `.json` config file?

hubcap/setup.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def build_config():
+    config = json.loads(os.environ['CONFIG'])
+    config['tracked_repos'] = {}
+
+    # necessary for as long as hubcap's git is disconnected from the github copy of the hub
+    response = requests.get('https://raw.githubusercontent.com/dbt-labs/hubcap/master/hub.json')
```
jtcohen6, line 33: I don't quite understand this. Why do we need to request the version of the `master` branch, rather than using the version available locally, and ensuring that Heroku is always running the latest version of the hubcap script (by pulling / checking out `master` earlier on)?

jtcohen6 (commented):
Took for a spin, not via Heroku (I don't have access to `dbt-hubcap-staging` ), just by running the script locally and manually inspecting the logs. Looking good!
```
2021-09-01 13:03:16 INFO     Push branches? True - ['bump-fivetran-dbt_stripe_source-1630493928', 'bump-fivetran-dbt_fivetran_utils-1630493928', 'bump-fivetran-dbt_youtube_analytics_source-1630493928', 'bump-calogica-dbt-expectations-1630493928', 'bump-jitsucom-dbt-jitsu-1630493928']
2021-09-01 13:03:16 INFO
2021-09-01 13:03:17 INFO     PR is already open for fivetran/stripe_source. Skipping.
2021-09-01 13:03:17 INFO     PR is already open for fivetran/fivetran_utils. Skipping.
2021-09-01 13:03:17 INFO     PR is already open for fivetran/youtube_analytics_source. Skipping.
2021-09-01 13:03:17 INFO     PR is already open for calogica/dbt_expectations. Skipping.
2021-09-01 13:03:17 INFO     PR is already open for jitsucom/jitsu. Skipping.
```

hubcap/setup.py
VersusFacit: We can but I'd rather just exclude these at the source frankly. This was a patch to get things work. I'd prefer to just block people from adding their packages that don't meet specifications rather than doing an exclusion dance. Is there a compelling reason that I may be missing for why we must keep everything in hub.json?

hubcap/setup.py
VersusFacit: Okay, haha, so this is the legacy code I referred to above. Since I've got the time today, I can edit this out. I'll do so in a few minutes and push a commit up so this will be resolved. 

For context, this was a necessary artifact from when the hubcap repo on Heroku was not the same as this hubcap repo. In fact, up until yesterday, the deployment involved manually copies of files, and to allow folks in the community to edit hub.json, this curl step had to be done separately. Now that we're auto deploying branches, we can take advantage of the local manifest (as we should).

VersusFacit: Patch (thankfully ended up being trivial). Things make more sense now.

hubcap/setup.py
VersusFacit: Did the thing and left a comment to remind we should just change the policy here probably.

VersusFacit:
Just reran this bad boy locally and on the staging cluster with the right results both times. (omg yay ENV variables making this easy).

cron.sh
```diff
@@ -1,10 +1,51 @@
...
+
+# ==
+# == Main
+# ==
+
+# specified in Heroku's config variables
+export ENV="${ENV-development}"
+if [ "$ENV" = 'prod' ] || [ "$ENV" = 'test' ]; then
+    email='drew@fishtownanalytics.com'
+    user_name='dbt-hubcap'
```
jtcohen6, line 35: Last comment, and absolutely something we can address later:

Do we want this to keep showing up as Drew? We have a ["machine user"](https://docs.github.com/en/developers/overview/managing-deploy-keys#machine-users) we could borrow for this purpose: @FishtownBuildBot

I think that could be as simple as:
```
    email='buildbot@fishtownanalytics.com'
    user_name='FishtownBuildBot'
```
And swapping whatever Drew-specific SSH key is in Heroku, for a personal access token tied to the BuildBot account

exclusions.json
```diff
@@ -0,0 +1,25 @@
+{
```
jtcohen6, line 1: In retrospect, I totally see your point: We could really just _remove_ these from `hub.json`, rather than having a totally separate `exclusions.json` to muddy the waters. Apologies for having made the additional ask; no need to make any changes for now.

jtcohen6 (approved):
Really exciting stuff. Nice work!!

cron.sh
VersusFacit: I'd like to address this later but yes, I have this in my Notion dev notes that are more expansive than what I've written here. Literally no disagreement and something I'd like to do as we continue to revamp these systems and add others.

exclusions.json
VersusFacit: All good. And I have this in my dev notes to eventually take care of this but frankly, the reason I'm not is that this will ultimately require telling people their packages have been delisted from hub.json (after they thought they had been added). To make this a smooth conversation, we need to provide a good contributing spec, which is something I believe I'll be tackling later this quarter as we think about revamping documentation and contribution guidelines.

PR #62, opened 2021-09-02
Need to update the Python runtime since it's stuck on 3.6.X.

VersusFacit:
image
Upgrading the "tech stack" in Heroku doesn't automatically change the Python version. We need to do that ourselves, per https://devcenter.heroku.com/articles/python-runtimes.


VersusFacit:
Deploy log is indeed specifying `3.9.0` and the build script is happy doing its thing as before (automagically on 3.9.7). 

image

image

PR #63, opened 2021-09-07
Hotfix - Replace Nones with empty iterables for package version specs

VersusFacit:
So, this [spec](https://github.com/dbt-labs/hub.getdbt.com/commit/ff1a01625ee55efee0370e26eb21b57b8dd4c8b5) for `dbtplyr` has a bug.  

The `null` on `packages:` should be a `[]` 'cus iterables. But, the `yaml.parse` function in python wants to return a `None` instead of an empty list. Then again, somewhere in hub (where I know not) wants an iterable, so we'll give it one.

### DIFF
--- a/hubcap/package.py
+++ b/hubcap/package.py
@@ -16,13 +16,15 @@ def has_dbt_project_yml(directory):
 
 def parse_pkg_name(repo_dir):
     with open(repo_dir / Path('dbt_project.yml'), 'r') as stream:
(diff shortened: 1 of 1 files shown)


VersusFacit:
Working PR [here](https://github.com/dbt-labs/hub.getdbt.com/pull/856). Bug fixed!!

Proof:
image

annafil (approved):

PR #74, opened 2021-09-17
Address the coming changes to Fivetran's default branches and package dependency versions

### LINKED ISSUE #44: Get package dependencies for specific release tag

### LINKED ISSUE #70: [Urgent] Bug: Build package configs from main branch (or master)
image

We need to ensure the hubcap script is prepared to handle packages with main, then prioritize that if there's no master branch. Currently, master is hardcoded into the script. We need main conditional logic that uses main then master when looking for commit shas to change into package specs. Otherwise, newer versions won't be added to hub.getdbt.com; the script will just break for any repo that does not retain a master branch.

Good news though! Because package specs depend only on commit shas, nothing has been changed for previous versions of packages.

This will partially address issue #69 .

VersusFacit:
Closes #70 
Closes #44 

## Purpose

Next week, Fivetran packages will begin building from main rather than master. The script is now powered to pull updates from whichever branch is default in a package project, whether it's master, main, or something entirely different--it's agnostic as to what the name actually is.

In order to do this, I began hacking away at code which I've been meaning to for at least a week. In unraveling a very nested for loop setup into decoupled function interfaces, I was able to find the right placement to address the long-dreaded package-version spec bug. Below is some proof.

image

The script generated this for an older version of dbt-expectations. [Here's what's currently checked into master of hub ](https://github.com/dbt-labs/hub.getdbt.com/blob/master/data/packages/calogica/dbt_expectations/versions/0.2.1.json) for this tag.

But, [here's what it should be](https://github.com/calogica/dbt-expectations/blob/0.2.0/packages.yml) -- same as the version generated by this new version of the script.

Moving forward, package readouts will be parsed from the packages.yml at the tag's specific commit sha. We've tried to fix this in the past with PR #45 but that didn't quite do the trick with the way the script was layered at the time. Drew proposed a different fix in a now closed PR. This is an attempt to mop up for good.

## Testing

It's mostly functions now which meant I could unit test as I went (and if wanted we can setup actual unit tests).

The script runs to completion as expected with no package updates to be made. It will generate a correct spec (technically one more correct than the production version of hubcap may run -- see the screenshot above) when there's at least one version to create.

## Side note

Fivetran's zuora repos are empty. This was causing bugs. I excluded them for now to get us moving.

Also, if you didn't notice, I move a decent amount of code around and began to employ the `gitPython` package which is quite decent today (even though old decade-old Stack overflow posts used to complain about it).

Lastly, to make things easier to debug, I added a command line option to preserve the commits and specs made throughout runtime so developers can peek more easily at what's going on under the hood.

### DIFF
--- a/cron.sh
+++ b/cron.sh
@@ -1,18 +1,41 @@
 #!/bin/bash
 
 # ==
+# == Parse args
+# ==
+
+POSITIONAL=()
+while [ "$#" != '0' ]; do
+  case "$1" in
+    # useful for wanting to save git-tmp and observe commits/build artifacts
+    -n|--no-cleanup)
+      CLEANUP='no'
+      shift
+      ;;
+    *)
+      POSITIONAL+=("$1")
+      shift
+      ;;
+  esac
+done
+set -- "${POSITIONAL[@]}"
+
+# ==
 # == Setup
 # ==
 
+# export to expose to python script later on
 export RUN_DIR="$(pwd)"
 export GIT_TMP="$RUN_DIR/git-tmp"
 
 exit_routine() {
   local -r exit_status="${1}"
 
+  if [ "${CLEANUP:-yes}" = 'yes' ]; then
     if [ -d "${GIT_TMP}" ]; then
         rm -rf "${GIT_TMP}"
     fi
+  fi
 
   exit "${exit_status}"
 }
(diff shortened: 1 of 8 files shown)


VersusFacit:
image

"Integration test" complete with a work `dbt deps`.

image

jasnonaz (approved):
Wow this is certainly a big refactor. Thank you for starting to untangle the rats nest that is HubCap.

I believe we can feel good about merging this after we tested live and it repulled in utils 0.7.3. 

A few follow on elements for us to start considering now that we have this ready to go:
1. Does the newly split up / functionalized script allow us to start integrating more tests in? 
2. Does this mean that we can now start thinking about switching the dbt maintained packages to use main instead of master.

Just some thoughts on next steps - to my review, this looks good to go

VersusFacit:
Yes to both!

PR #77, opened 2021-10-29
Bonus code rewrites

VersusFacit:
@nathaniel-may Gave me some awesome advice for improving the code further. I took those suggestions to heart and reduce side effects in several functions. Other improvements revolved around getting rid of cumbersome data structures and adopting classes for encapsulating behaviors. The general sum is as follows:

- add intent-based comments for functions _and_ modules
- most functions, especially early on either have no side effects or are functions without a return value but explicitly changing machine state
- removing some of the more complicated dictionaries in dictionaries that were absolutely insane to work with
- add raise_for_requests (in lieu of exceptions) to catch more bugs
- general bugfixes that occasionally happened with unlucky indexing
- I discovered a bug that would have come up in the future, namely that if two organizations named their package the same thing (this could happen!) the script would fail -- now packages are considered by org/pkg name pairs so no collisions can occur
- list comprehensions (without side effects) adopted to get rid of nasty for loops
- the script will commit using dbt-hubcap as the git user BUT this will not corrupt local git repos which local runs won't mess up users' computers

This represents the last major rewrite I'll be doing on this repository in the near future. Not to have this PR drag on more than a handful of days, I'm forgoing other opportunities to optimize code style.

This needs a solid integration test before merging. We'll do it with a new stable version and a beta release of some package, likely a lesser dbt one we can control well.

### DIFF
--- a/cron.sh
+++ b/cron.sh
@@ -43,9 +43,9 @@ exit_routine() {
     function reset_git_params() {
       # reset machine's git config state to pre-script invocation
       if [ -z "${2}" ]; then
-          git config --local --unset user."${1}"
+          git config --global --unset user."${1}"
       else
-          git config --local user."${1}" "${2}"
+          git config --global user."${1}" "${2}"
       fi
     }
     reset_git_params 'email' "$PRIOR_GIT_EMAIL"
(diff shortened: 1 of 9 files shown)


VersusFacit:
@nathaniel-may Just cc-ing you that this is the culmination of your suggestions. I broke up some other nested logic by way of a Class. Oh boy, logic and behavior packaged for ease of use. Thanks for your eyes on this 🙌

PR #79, opened 2021-11-12
Update hub.json

aneiderhiser:
Tuva Health is building a new set of dbt packages and this is the first one we'd like to add to dbt hub.

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -160,5 +160,8 @@
     ],
     "avohq": [
         "avo_audit"
+    ],
+    "tuva-health": [
+        "staging"
     ]
 }


VersusFacit:
Looks like you've got the project prerequisites. Kudos on your first package! 🎉  I look forward to adding more

Our production server should pick this up on the hour. I'll be watching the logs personally (I'm monitoring the health of some recent codepushes).

VersusFacit:
I'm really sorry to do this to you, but after looking deeper into your models, I noticed some [hardcoded model names](https://github.com/tuva-health/staging/blob/main/models/diagnoses.sql) in your package.

For example, in this model, you hardcode a public schema table. That can result in problems for users who may not have that schema configured. Instead, things like dbt's ref function should be put to use. I've got an arbitrary example from a [Fivetran package](https://github.com/fivetran/dbt_github/blob/main/models/github__issues.sql).

Apologies for the back and forth here. Please do ask question and let's get this added after seeing the code adjusted a touch :)

aneiderhiser:
We need a configuration layer in our package i.e. where users map in their source data.  Is the better way to do this to define source tables using variables and then feed these into stage?

PR #82, opened 2021-11-17
Add tuva package by tuva-health

aneiderhiser:


aneiderhiser:
@jtcohen6 Just wanted to check and confirm that I'm still waiting on your review before I can merge and start on the next steps to add my project to dbt hub packages.  Thanks.

VersusFacit:
Hi there, and apologies for having you blocked here. Let's untangle that.

So, in files like `staging/models/encounters.sql` you have hard-coded `from hcup.public.encounters`. As I understand it from your prior messages, you'd like to have users transformation the data into a sort of base table setup that you then provide these transforms on top of.

Users may not have this schema setup available or may prefer something else. But, we can't use ref since the user may not want to have that exact name setup in their project. Moreover, they can't edit these files locally since every `dbt deps`, those files will be overwritten or at the very least in conflict with what's in the main branch of your repo.

Here's a cool way to template all this out which provides your users a way to dynamically alter table names. I'd suggest dropping this in for each table which expects to draw from some existing base table:

Change lines like `from hcup.public.encounters` to `{{ var('tuva_encounters') }}` and encouraging users to add lines this in their project yaml or specific schema defined property files:

```yaml
vars:
    encounters_base: ref('<users name for transformed base table>')
```

Please ask any questions. I'll keep an eye on the PR. Special Jinja-dbt functions are documented on our site!

aneiderhiser:
I'm trying to push a different repo to your hub - I believe I've fixed the issue you are talking about in this new repo.  Check out the repo here: https://github.com/tuva-health/tuva

VersusFacit:
Thanks for clarifying! Pardon me, you're correct about having a different repo from the one I originally linked. But for both, I was `{{ source('source','encounters') }}`. The preferred strategy is using `var` which then templates out a user defined `ref` in a user-controlled yaml.

Hence, why I linked the:
```yaml
vars:
    encounters_base: ref('<users name for transformed base table>')
```
which can then be referenced like so in your package's models:

```sql
select *
from {{ var('user_defined_table_name') }}
```
You should tell people in the docs for your package what the default name could be, but encourage users to override it, as they may name their table any number of things based on their setup. 

That way, when you compile all this, dbt will replace the `var` with a `ref` and then that `ref` will be replaced with the user's actual table.

VersusFacit:
Some relevant documentation includes:
* [docs](https://docs.getdbt.com/reference/dbt-jinja-functions/var)
* [user defined property files] https://docs.getdbt.com/reference/configs-and-properties

aneiderhiser:
Okay - this is very helpful thank you!

What's the best way to configure this to allow the user's source data to reside outside the database and/or schema the package is writing to?  Is the variable intended to include only the table name e.g.:

vars:
    encounters_base: "ref('<table_name>')"

or is it intended for the user to define the fully-specified table name w/ database and schema, e.g.:

vars:
    encounters_base: "ref('<database_name.schema_name.table_name>')"

If it's the former, what's the best way for the user to specify their database and schema name for their source data?

The way I have it configured currently (using variables in sources) allows for this.  I would like to allow for this same flexibility under the recommended approach (the approach you're proposing).

aneiderhiser:
@VersusFacit I updated our project based on your recommendation.  I'm using vars in the config like you suggest for the table names.  However I'm still using sources to specify how these tables are referenced.  Please let me know if this approach is OK.  Here's the project link for convenience: https://github.com/tuva-health/tuva

VersusFacit:
> What's the best way to configure this to allow the user's source data to reside outside the database and/or schema the package is writing to? Is the variable intended to include only the table name

My suggestion is that all your tables be like `{{ var('encounters_base') }}` in models. Then, users can either use `source` (to reference extra-schema models) or `ref` (to use intra-schema models). That's the pattern we encourage on our website and in our docs, so you can document this for your users or trust them to know the paradigm.

So explicitly, `from {{ source('source',var('encounters_source')) }}` should be `from {{ var('encounters_source') }}`, then we trust users to know that in their yamls then can use a `source` or `ref` directive.

I'll check tomorrow so we can unblock you on this. Sorry for the back and forth, but we're quite close on getting this solid 💪

VersusFacit:
If you'd like, you can message me directly on the dbt slack (I'm just under Mila) over there. Would be happy to give some expedited feedback there tomorrow.

aneiderhiser:
Sounds great - I think implementing it exactly as you suggest, i.e.  from {{ source('source',var('encounters_source')) }} is the from statement in all the staging tables.  Are there any other changes you'd like me to make?

VersusFacit:
Hi there! Circling back to this. In [this PR](https://github.com/dbt-labs/hubcap/pull/86), I added this package, and hence I will close out this PR. In the future, when you have more packages to add, please add them to the repo directly. We're so happy you stuck with us and wish you happy developing.

Thank you for your patience and I look forward to seeing other packages end up on the docket.

PR #85, opened 2021-12-14
Add Matatika dbt-tap-googleads

DanielPDWalker:
Hi, could we please add our dbt-tap-googleads project to the hub.

We test this project daily end to end in our app. Behind the scenes this creates a Meltano project, adds a singer-tap, our dbt-tap-googleads project, and some datasets. Does a complete ETL and then we assert that our datasets, which work off the dbt models, contain data. 

This is the second PR we have, whichever gets merged in first we'll update the other PR.

Let me know if there is anything else you want to know or want me to change :)

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -173,5 +173,8 @@
     ],
     "avohq": [
         "avo_audit"
+    ],
+    "Matatika": [
+        "dbt-tap-googleads"
     ]
 }


joellabes:
Hi @DanielPDWalker, thanks for opening this! I've had a quick look and it looks good. 

My only piece of feedback is that by referencing tables as `{{ schema }}.stream_adgroupsperformance`, your models won't participate in the dbt DAG. It could be better to add them as sources. (I also noticed that [campaign is misspelled here](https://github.com/Matatika/dbt-tap-googleads/blob/master/models/base/ad_groups_performance.sql#L9) 🙊 )

Happy to merge this and you can make other changes by cutting a new version of the package 🎉

PR #88, opened 2022-01-03
Remove dbt-labs/facebook_ads from exclusions.json

joellabes:
As discussed with @VersusFacit to solve the updated packages not being pulled through to Hub
image


VersusFacit:
LGTM

PR #101, opened 2022-04-04
Add Matatika dbt-tap-spotify

DanielPDWalker:
Hey, could we please get our `dbt-tap-spotify` package into the hub.

For testing, alongside the in project tests, we test these projects daily end to end in our app.  Behind the scenes this creates a Meltano project, adds a singer-tap, our dbt-tap project, and some datasets. Does a complete ETL and then we assert that our datasets, which work off the dbt models, contain data. This happens for both snowflake and postgres.

Let me know if there is anything else wanted/needed :)

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -184,7 +184,8 @@
         "dbt-tap-solarvista",
         "dbt-tap-meltano",
         "dbt-tap-trello",
-        "dbt-tap-google-analytics"
+        "dbt-tap-google-analytics",
+        "dbt-tap-spotify"
     ],
     "tuva-health": [
         "tuva"


dbeatty10 (approved):
Thanks for submitting this package to the Hub! Will be included in our next release (today).

Curious: noticed that `DBT_TARGET_SCHEMA` is used within the tests rather than `{{ source() }}`. Was there some reason that `{{ source() }}` wouldn't work for you within your tests?

DanielPDWalker:
@dbeatty10 Completely forgot about the `DBT_TARGET_SCHEMA` in the tests. Will be looking to update them and making a new release soon!

PR #109, opened 2022-04-29
Added Github DBT package for Airbyte

milo157:
Added a DBT package for the Github connector on DBT

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -25,6 +25,9 @@
         "dbt-external-tables",
         "dbt_metrics"
     ],
+    "cerebriumai": [
+        "airbyte_dbt_github"
+    ],
     "fivetran": [
         "dbt_netsuite",
         "dbt_netsuite_source",


dbeatty10 (approved):

dbeatty10:
Looks good @milo157 !

One comment: best not to pin your utils dependency to 0.8.4 in [packages.yml](https://github.com/CerebriumAI/airbyte_dbt_github/blob/e5b73f52dab7552b52ff1d243d2b2c97c3f48ce2/packages.yml#L3) - it should be any 0.8.x version so `version: [">=0.8.0", "<0.9.0"]` instead.

It will show up on [hub.getdbt.com](https://hub.getdbt.com/) after our next release of that website.

dbeatty10:
You're up on hub.getdbt.com now:
https://hub.getdbt.com/cerebriumai/

milo157:
Awesome! Thanks @dbeatty10! Quick question, if I change the package name in the main repo will it automatically update?

dbeatty10:
If you change the package name, it will automatically create a new entry on hub.getdbt.com.

Then I'll do a linkage behind the scenes so that it hides the old name. Anyone that uses the old name will be automatically re-directed to the new name.

I'm guessing you're asking because these are the same (just old name vs. new)?
image

milo157:
Correct. If you can remove the plain GitHub name. The latest should be airbyte_github

Thank you!

dbeatty10:
✅ Done!

Thanks for reaching out about this so we could get it all squared away.

PR #110, opened 2022-05-02
add dbt-client

gabriel-milan:


dbeatty10:
Hi @gabriel-milan -- this is a cool JSON-RPC client for dbt! :heart_eyes: 

Currently, the best place to tell others about this would be the [#i-made-this](https://getdbt.slack.com/archives/C01NH3F2E05) channel in the [dbt Community Slack](https://www.getdbt.com/community/join-the-community/).

We can't include it on [hub.getdbt.com](https://hub.getdbt.com/) at this time -- it can only include those meant to be installed with `dbt deps`. (That's why the readme says the part about "your project must have a dbt_project.yml".)

We've got aims for the future to create a place to show off and share cool tools like this.

PR #112, opened 2022-05-03
Update hub.json

fivetran-joemarkiewicz:
Adding the following packages to the hub site in preparation of initial release:
- dbt_google_play
- dbt_google_play_source
- dbt_apple_store
- dbt_apple_store_source
- dbt_app_reporting

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -106,7 +106,12 @@
         "dbt_linkedin_pages_source",
         "dbt_social_media_reporting",
         "dbt_sage_intacct",
-        "dbt_sage_intacct_source"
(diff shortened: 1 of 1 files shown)


dbeatty10 (commented):
Looks good to me, Joe!

I think this will put you over 80(!) packages on hub.getdbt.com 🤯 

I just need to check first what will happen if we merge this before each package has at least one tag (with the version number).

dbeatty10 (approved):
👍 Apparently it should be fine even if there's no tags yet.

PR #116, opened 2022-05-13
Added Stripe and Intercom packages

milo157:


hub.json
```diff
@@ -28,7 +28,9 @@
     "cerebriumai": [
         "airbyte_dbt_github",
         "airbyte_dbt_google_ads",
-        "airbyte_dbt_shopify"
+        "airbyte_dbt_shopify",
+        "airbyte_dbt_intercom",
+        "airbyte_dbt_stripe"
```
dbeatty10, line 33: I see the intercom package in the list of public repos on GitHub. But I'm not seeing the stripe one. Is it possible is it categorized as private rather than public?

Here's where I'm looking, and I only see 8 public packages:
https://github.com/orgs/CerebriumAI/repositories

dbeatty10 (changes requested):
Could you check if airbyte_dbt_stripe is a public repo?

milo157:
Sorry forgot to make public! All good now.

dbeatty10:
Cool, I can see both repos now 👍 

One more thing:
* you'll need to add a tag to each repo so that the version numbers will show up in the drop down menus like [here](https://hub.getdbt.com/cerebriumai/airbyte_github/latest/).

The Hub gets version numbers from tags rather than from `dbt_project.yml`

milo157:
Done. Sorry we got sloppy with this one.

dbeatty10 (approved):
> Done. Sorry we got sloppy with this one.

No worries at all -- lots of little details, and I'm there for you!

PR #117, opened 2022-05-23
Added facebook ads and Pipedrive dbt packages

milo157:
Added Facebook Ads and Pipedrive DBT packages for the Airbyte connectors


dbeatty10 (approved):
👍

PR #126, opened 2022-06-29
Update hub.json to add carbon analytics package

IL-NINA:
Add new carbon analytics package


dbeatty10 (approved):
Woot! 🎉 Very excited about this @IL-NINA 🌍 

This should show up for the first time at [hub.getdbt.com](https://hub.getdbt.com/) in about an hour.

PR #130, opened 2022-07-22
Make hyphen optional for prerelease suffix

### LINKED ISSUE #129: Support prerelease identifiers without hyphens
Prompted by @NiallRees's exciting `1.0.0b1` release of `dbt_artifacts`: https://github.com/brooklyn-data/dbt_artifacts/releases/tag/1.0.0b1

https://github.com/dbt-labs/hubcap/blob/d82a8a284748886eb28bf45521e76c6fb6a1adbe/hubcap/version.py#L14-L15

I think that's SemVer-official, but Python/pip actually doesn't support the hyphen. [From PEP 440](https://peps.python.org/pep-0440/#:~:text=Semantic%20versions%20containing%20a%20hyphen%20(pre%2Dreleases%20%2D%20clause%2010)%20or%20a%20plus%20sign%20(builds%20%2D%20clause%2011)%20are%20not%20compatible%20with%20this%20PEP%20and%20are%20not%20permitted%20in%20the%20public%20version%20field.):
> Semantic versions containing a hyphen (pre-releases - clause 10) or a plus sign (builds - clause 11) are not compatible with this PEP and are not permitted in the public version field.

So, it's pretty common for folks to use the hyphenless prerelease identifier, even if it's not "real" semver. The Core team stumbled across this inconsistency a few months ago (https://github.com/dbt-labs/dbt-core/issues/4741), and decided to let it be. The `dbt-core` semver logic supports both:
```
>>> from dbt.semver import VersionSpecifier
>>> VersionSpecifier.from_version_string("1.0.0b1")
VersionSpecifier(major='1', minor='0', patch='0', prerelease='b1', build=None, matcher=<Matchers.EXACT: '='>)
>>> VersionSpecifier.from_version_string("1.0.0-b1")
VersionSpecifier(major='1', minor='0', patch='0', prerelease='b1', build=None, matcher=<Matchers.EXACT: '='>)
```

IMO:
- We should aim for consistency between Hubcap + dbt-core (though it's a very good thing that Hubcap removed its dependency on dbt-core!)
- We should accept both `1.0.0b1` and `1.0.0-b1` as valid semantic version identifiers, with prerelease suffix b1.
- It's just a matter of adding one little qmark (`?`) to the regex string

jtcohen6:
resolves #129

Not sure how to test this! It's been a little while since I did anything in this repo :)

### DIFF
--- a/hubcap/version.py
+++ b/hubcap/version.py
@@ -12,7 +12,7 @@ from cmd import *
 def parse_semver_tag(tag):
     '''use regexes to parse the semver tag into groups'''
     # regex taken from official SEMVER documentation site
-    match = re.match('^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$',
+    match = re.match('^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-?(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$',
         tag[1:] if tag.startswith('v') else tag
     )
     return match


dbeatty10 (commented):
I'm going to work on #131 (and #132) so that we can try to add some tests here before merging.

dbeatty10:
## Testing framework
#133 adds some simple regex tests and instructions for running them. Currently, they are crafted strictly towards SemVer 2.0, so the unit tests would need to be adjusted slightly to accommodate an optional hyphen for prerelease suffix.

## Can we have it all?
Is there any way we could achieve consistency between Hubcap + db-core _and_  have strict SemVer for dbt packages _and_ follow PEP 440 version strings for dbt adapters?

[Converting versions between PyPI and semver](https://python-semver.readthedocs.io/en/latest/advanced/convert-pypi-to-semver.html#converting-versions-between-pypi-and-semver) provides some useful commentary and code. You can see how the `semver` project's [git tags](https://github.com/python-semver/python-semver/tags) compare to their [PyPI releases](https://pypi.org/project/semver/#history):

| git tag     | PyPI release |
|-------------|--------------|
| 3.0.0-dev.3 | 3.0.0.dev3   |
| 3.0.0-dev.2 | 3.0.0.dev2   |
| 3.0.0-dev.1 | 3.0.0.dev1   |
| 2.13.0      | 2.13.0       |
| 2.12.0      | 2.12.0       |
| 2.11.0      | 2.11.0       |

hubcap/version.py
```diff
@@ -12,7 +12,7 @@
 def parse_semver_tag(tag):
     '''use regexes to parse the semver tag into groups'''
     # regex taken from official SEMVER documentation site
```
joellabes, line 14: Does this regex now slightly deviate from the official site? Or had it moved forward and we hadn't moved with it?

jtcohen6:
> [Converting versions between PyPI and semver](https://python-semver.readthedocs.io/en/latest/advanced/convert-pypi-to-semver.html#converting-versions-between-pypi-and-semver) provides some useful commentary and code.

It also makes me wonder if we should be using these packages, rather than our own regex:
```python
from semver import Version as SemVersion
from packaging.version import Version as PyPIVersion

def parse_semver_tag(tag):
    version = SemVersion(tag) or PyPIVersion(tag) # probably needs to be try/except
    return str(version)
```

That's a bigger lift than what we're trying to unblock in this PR.

hubcap/version.py
dbeatty10: The regex in the `main` branch currently matches the "named groups" regular expression in [SemVer 2.0.0](https://semver.org/spec/v2.0.0.html#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string) precisely:
`^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`

That regex has not changed (and probably will not change within SemVer 2.x).

This PR just adds a single character change that allows for the dash prefix for pre-releases to be optional:
Screen Shot 2022-07-26 at 9 08 31 AM

This would to allow for dbt package versions to have greater overlap with [PEP 440](https://peps.python.org/pep-0440/) even though SemVer 2.0.0 doesn't allow it to be optional.

joellabes: Yeah so my point is that if it's no longer the official SEMVER regex, we should update the comment to say something like "based on the SEMVER regex but with tweaks for PEP compatibility"

PR #104, opened 2022-04-11
tuva health package updates

aneiderhiser:
changed "tuva" to "core" and added "chronic_conditions" and "readmissions" repos.

PR #134, opened 2022-07-25
Updated ID Resolution package from RudderLabs

DipanjanAtRudderLabs:


hub.json
```diff
@@ -44,7 +44,7 @@
     ],
 	"rudderlabs": [
 		"dbt-sessionization",
-		"dbt-id-resolution",
+		"dbt-id-stitching",
```
joellabes, line 47:
```suggestion
		"dbt-id-stitching",
		"dbt-id-resolution",
```

joellabes (changes requested):
@DipanjanAtRudderLabs the licence file should be bundled with your project and/or referenced from your README, not added to the dbt hub - this does not contain any code on its own, it's just a pointer to where the code actually is.

Upon further digging, I also think that for your legacy package to remain available on the Hub, it needs to stay in `hub.json` but be added to [blocklist.json](https://github.com/dbt-labs/hub.getdbt.com/blob/master/data/blocklist.json) on the hub.getdbt.com repo. (We will do that on our end once this changeset is merged).

hub.json
dbeatty10: @joellabes Mea culpa -- I recommended the swap to them.

Removing from `hub.json` has worked for me before ([code diff](https://github.com/dbt-labs/hubcap/pull/104/files), [org after](https://hub.getdbt.com/tuva-health/),  [old package after](https://hub.getdbt.com/tuva-health/tuva/latest/)). I think this is because `hubcap.py` is append-only -- I don't think it removes anything after it's been merged to [docs.getdbt.com](https://github.com/dbt-labs/docs.getdbt.com).

### Guidance given

For full context, here's the guide I gave the RudderStack team over Slack:

Here's what you'll do:
1. Create [at least one tag](https://github.com/rudderlabs/dbt-id-stitching/tags) for [dbt-id-stitching](https://github.com/rudderlabs/dbt-id-stitching)
1. Submit a PR similar to [this one](https://github.com/dbt-labs/hubcap/pull/104/files) that replaces your old repo name with your new one [here](https://github.com/dbt-labs/hubcap/blob/d82a8a284748886eb28bf45521e76c6fb6a1adbe/hub.json#L135)

Then myself or @joellabes will do these actions:
1. Review and merge your PR from 2) above
1. Make sure the new package shows up on the [dbt package Hub](https://hub.getdbt.com/)
1. Do some code changes on our end to handle "archiving" the [rudder_id_resolution](https://hub.getdbt.com/rudderlabs/rudder_id_resolution/latest/) package


### Rationale for guidance given
In a Slack convo, the RudderStack team emphasized that they wanted to **completely replace** the old with the new. But...the new package is a complete re-write within a different repo, and I wanted to be sure that any existing unpinned versions of the old package in `packages.yml` aren't redirected to the new one (which would definitely break).

I recommended the swap to them due to the following beliefs:
- `dbt deps` will continue to work for all references to the old package
- the hubcap.py script will stop tracking tagged versions of the old package
- the old package will no longer display on the main Hub website or the org listing
- but the old package will still [keep a stand-alone page](https://hub.getdbt.com/rudderlabs/rudder_id_resolution/latest/) (without a redirect)

I might be overly cautious due to misunderstanding how the redirect property + hubcap.py + `dbt deps` works. I might have also just given sub-optimal guidance.

joellabes: I tried installing `https://hub.getdbt.com/dbt-labs/shopify/latest/` and `https://hub.getdbt.com/fivetran/dbt_zuora/latest` through the Hub and they both came back as unavailable, which is why I got worried. 

I just tried `https://hub.getdbt.com/fishtown-analytics/shopify/latest/` and that did work (Fishtown instead of dbt Labs). The Fivetran one should have remained consistently available though, given that they haven't renamed 🤷 

Looks like you've done more research so I'm willing to follow your lead 👌  

In that case @DipanjanAtRudderLabs, the only thing to do is remove the licence.md and we can get this merged!

DipanjanAtRudderLabs:
Closing this PR to create a new one based on recent discussions

PR #137, opened 2022-07-26
Updated ID Resolution DBT package

DipanjanAtRudderLabs:
Modified PR to include only hub.json changes


dbeatty10:
Hi @DipanjanAtRudderLabs this PR can't be merged as-is due to merge conflicts.

Your branch diverged from this repository's default branch after 2020-07-29 and there's been quite a few additions since then :)

There are multiple ways to solve this, but the most straight-forward would be for you to completely start over:
1. Re-clone https://github.com/dbt-labs/hubcap from scratch
1. Make your one-line change
1. Submit a new PR

The alternatives to resolve this would be more complicated and involved -- not worth it for a one-line change. As a bonus, starting over will make the commits related to the LICENSE file disappear too.

dbeatty10 (changes requested):
See comment above for suggested path forward.

PR #139, opened 2022-07-28
Package best practices doc

joellabes:
Based on the lessons learned from adding a bunch of packages to the Hub, here's a spec for would-be contributors to validate against before opening a hubcap PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,4 +14,4 @@ Assorted constraints:
 ### Adding your package to hubcap
 Currently, only packages hosted on a GitHub repo are supported.
 
-To add your package, open a PR that adds your repository to [hub.json](hub.json).
+To add your package, open a PR that adds your repository to [hub.json](hub.json). A dbt Labs team member will review your PR and provide a cursory check of your new package against [best practices](package-best-practices.md).
--- a/package-best-practices.md
+++ b/package-best-practices.md
@@ -25,19 +25,22 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     - MAY assume table names, particularly if the package was built to support tables created by a known tool.
 
 ### Dependencies
+#### Dependencies on dbt Core
 - Packages requiring a minimum version of `dbt-core` MUST declare it in the `require-dbt-version` property of `dbt_project.yml`.
-- Packages requiring a minimum version of `dbt-core` SHOULD declare compatibility with all subsequent minor and patch releases of that major version. 
+- Packages requiring a minimum version of `dbt-core` SHOULD allow all subsequent minor and patch releases of that major version. 
     - For example, a package which depends on functionality added in dbt Core 1.2 SHOULD set its `require-dbt-version` property to `[">=1.2.0", "<2.0.0"]`.
-Packages with a dependency on another package:
-    - MAY, if the dependent package is major-version-zero, pin to the current minor version only.
-    - SHOULD NOT pin to a patch version of the dependent package unless they are aware of an incompatibility between their package and the dependency.
-    - SHOULD reference the version in the dbt Package Hub when available, as opposed to a `git` installation.
+#### Dependencies on other packages defined in `packages.yml`:
+- Packages SHOULD import their dependencies from the dbt Package Hub when available, as opposed to a `git` installation.
+- Packages SHOULD NOT pin to a patch version of their imported package unless they are aware of an incompatibility.
+- Packages importing another package which has reached major version one or later SHOULD allow all subsequent minor and patch releases of that major version. 
+- Packages importing another package which is major version zero MAY pin to the current minor version only.
 
 ### Interoperability
 - Packages MUST NOT override dbt Core behaviour in such a way as to impact other dbt resources (models, tests, etc) not provided by the package.
+- Packages MUST use the cross-database macros built into dbt Core where available, such as `{{ datediff }}` and `{{ type_string() }}`.
 - Packages SHOULD disambiguate their resource names to avoid clashes with nodes that are likely to already exist in a project. 
     - For example, packages SHOULD NOT provide a model simply called `users`.
 
 ### Releases and updates
+- Packages' git tags MUST validate against the regex defined in [version.py](/hubcap/version.py).
 - Packages SHOULD follow the guidance of the [Semantic Versioning Specification](https://semver.org/).
-- Packages MUST create a git tag with the version number that follows the accepted version format.
--- a/package-best-practices.md
+++ b/package-best-practices.md
@@ -1,0 +1,41 @@
+# Best practices when submitting a package for public use
+
+Packages are a key part of the dbt experience. Practitioners and vendors alike can [contribute to the knowledge loop](https://github.com/dbt-labs/corp/blob/main/values.md#we-contribute-to-the-knowledge-loop) by enabling compelling, batteries-included use cases.
+
+There are hundreds of packages on the dbt Package Hub already, and over time we have identified a handful of common mistakes that new package authors make as they take a self-contained package and prepare it for publication. We have listed these below to help you and your users get the best experience.
+
+Although the dbt Labs team completes a cursory review of new packages before adding them to the Package Hub, ultimately you are responsible for your package's performance.
+
+To avoid ambiguity, the bullets below are pretty formal, but we are a friendly bunch! If you need help preparing a package for deployment to the dbt Package Hub, reach out in [#package-ecosystem](https://getdbt.slack.com/archives/CU4MRJ7QB/) in the [dbt Community Slack](https://getdbt.com/community).
+
+---
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+### First run experience
+- Packages SHOULD contain a README file which explains how to get started with the package and customise its behaviour.
+
+### Customisability 
+- Packages MUST NOT hard-code table references, and MUST use `ref` or `source` instead.
+- Packages MAY provide a mechanism (such as variables) to enable/disable certain subsets of functionality.
+- Packages for data transformation:
+    - SHOULD provide a mechanism (such as variables) to customise the location of source tables.
+    - SHOULD NOT assume database/schema names in sources. 
+    - MAY assume table names, particularly if the package was built to support tables created by a known tool.
+
+### Dependencies
+- Packages requiring a minimum version of `dbt-core` MUST declare it in the `require-dbt-version` property of `dbt_project.yml`.
+- Packages requiring a minimum version of `dbt-core` SHOULD declare compatibility with all subsequent minor and patch releases of that major version. 
+    - For example, a package which depends on functionality added in dbt Core 1.2 SHOULD set its `require-dbt-version` property to `[">=1.2.0", "<2.0.0"]`.
+Packages with a dependency on another package:
+    - MAY, if the dependent package is major-version-zero, pin to the current minor version only.
+    - SHOULD NOT pin to a patch version of the dependent package unless they are aware of an incompatibility between their package and the dependency.
+    - SHOULD reference the version in the dbt Package Hub when available, as opposed to a `git` installation.
+
+### Interoperability
+- Packages MUST NOT override dbt Core behaviour in such a way as to impact other dbt resources (models, tests, etc) not provided by the package.
+- Packages SHOULD disambiguate their resource names to avoid clashes with nodes that are likely to already exist in a project. 
+    - For example, packages SHOULD NOT provide a model simply called `users`.
+
+### Updates
+- Packages SHOULD follow the guidance of the [Semantic Versioning Specification](https://semver.org/).
--- a/package-best-practices.md
+++ b/package-best-practices.md
@@ -40,7 +40,6 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     - For example, packages SHOULD NOT provide a model simply called `users`.
 
 ### Releases and updates
-### Releases and updates
 - Packages' git tags MUST validate against the regex defined in [version.py](/hubcap/version.py).
 - Packages SHOULD follow the guidance of [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html).
     - Note in particular the recommendation for production-ready packages to be version `1.0.0` or above
--- a/package-best-practices.md
+++ b/package-best-practices.md
@@ -43,4 +43,4 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 ### Releases and updates
 - Packages' git tags MUST validate against the regex defined in [version.py](/hubcap/version.py).
-- Packages SHOULD follow the guidance of the [Semantic Versioning Specification](https://semver.org/).
+- Packages SHOULD follow the guidance of [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html).
--- a/package-best-practices.md
+++ b/package-best-practices.md
@@ -14,6 +14,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 ### First run experience
 - Packages SHOULD contain a README file which explains how to get started with the package and customise its behaviour.
+- The documentation SHOULD indicate which data warehouses are expected to work with this package.
 
 ### Customisability 
 - Packages MUST NOT hard-code table references, and MUST use `ref` or `source` instead.
@@ -37,5 +38,6 @@ Packages with a dependency on another package:
 - Packages SHOULD disambiguate their resource names to avoid clashes with nodes that are likely to already exist in a project. 
     - For example, packages SHOULD NOT provide a model simply called `users`.
 
-### Updates
+### Releases and updates
 - Packages SHOULD follow the guidance of the [Semantic Versioning Specification](https://semver.org/).
+- Packages MUST create a git tag with the version number that follows the accepted version format.
--- a/package-best-practices.md
+++ b/package-best-practices.md
@@ -31,13 +31,11 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
     - For example, a package which depends on functionality added in dbt Core 1.2 SHOULD set its `require-dbt-version` property to `[">=1.2.0", "<2.0.0"]`.
 #### Dependencies on other packages defined in `packages.yml`:
 - Packages SHOULD import their dependencies from the dbt Package Hub when available, as opposed to a `git` installation.
-- Packages SHOULD NOT pin to a patch version of their imported package unless they are aware of an incompatibility.
-- Packages importing another package which has reached major version one or later SHOULD allow all subsequent minor and patch releases of that major version. 
-- Packages importing another package which is major version zero MAY pin to the current minor version only.
-
+- Packages SHOULD specify the widest possible range of supported versions, to minimise issues in dependency resolution. 
+    - In particular, packages SHOULD NOT pin to a patch version of their imported package unless they are aware of an incompatibility.
 ### Interoperability
 - Packages MUST NOT override dbt Core behaviour in such a way as to impact other dbt resources (models, tests, etc) not provided by the package.
-- Packages MUST use the cross-database macros built into dbt Core where available, such as `{{ datediff }}` and `{{ type_string() }}`.
+- Packages intended for use with multiple platforms MUST use the cross-database macros built into dbt Core where available, such as `{{ except() }}` and `{{ type_string() }}`.
 - Packages SHOULD disambiguate their resource names to avoid clashes with nodes that are likely to already exist in a project. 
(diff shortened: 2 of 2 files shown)


package-best-practices.md
```diff
@@ -0,0 +1,41 @@
...
+    - SHOULD provide a mechanism (such as variables) to customise the location of source tables.
+    - SHOULD NOT assume database/schema names in sources. 
+    - MAY assume table names, particularly if the package was built to support tables created by a known tool.
+
+### Dependencies
+- Packages requiring a minimum version of `dbt-core` MUST declare it in the `require-dbt-version` property of `dbt_project.yml`.
+- Packages requiring a minimum version of `dbt-core` SHOULD declare compatibility with all subsequent minor and patch releases of that major version. 
+    - For example, a package which depends on functionality added in dbt Core 1.2 SHOULD set its `require-dbt-version` property to `[">=1.2.0", "<2.0.0"]`.
+Packages with a dependency on another package:
+    - MAY, if the dependent package is major-version-zero, pin to the current minor version only.
```
joellabes, line 31: Does "dependent" make clear that it is the package that the main package is importing? I think it does until I think too hard about dependents in the real world normally being children, which would make them the downstream one as opposed to looking upstream. Not sure if I'm too in my head or not, and don't have a better word.

dbeatty10: Would "requirement" be a less confusing word?

An additional possibility: the "Dependencies" could be split into two parts:
- compatible versions of dbt Core
- compatible versions of package requirements

The first bullet point references `dbt_project.yml` -- maybe there could be one could referencing `packages.yml` also?

- Each package requirement MUST be declared in `packages.yml`.
- Requirement ranges SHOULD be as wide as compatibility allows.

☝️ might need some further word-smithing

README.md
```diff
@@ -14,4 +14,4 @@ Assorted constraints:
 ### Adding your package to hubcap
 Currently, only packages hosted on a GitHub repo are supported.
 
-To add your package, open a PR that adds your repository to [hub.json](hub.json).
+To add your package, open a PR that adds your repository to [hub.json](hub.json). A dbt Labs team member will review your PR and provide a cursory check of your new package against [best practices](package-best-practices.md).
```
dbeatty10, line 17: 😍

package-best-practices.md
```diff
@@ -0,0 +1,41 @@
+# Best practices when submitting a package for public use
+
+Packages are a key part of the dbt experience. Practitioners and vendors alike can [contribute to the knowledge loop](https://github.com/dbt-labs/corp/blob/main/values.md#we-contribute-to-the-knowledge-loop) by enabling compelling, batteries-included use cases.
```
dbeatty10, line 3: 🎯

package-best-practices.md
```diff
@@ -0,0 +1,41 @@
...
+
+There are hundreds of packages on the dbt Package Hub already, and over time we have identified a handful of common mistakes that new package authors make as they take a self-contained package and prepare it for publication. We have listed these below to help you and your users get the best experience.
+
+Although the dbt Labs team completes a cursory review of new packages before adding them to the Package Hub, ultimately you are responsible for your package's performance.
+
+To avoid ambiguity, the bullets below are pretty formal, but we are a friendly bunch! If you need help preparing a package for deployment to the dbt Package Hub, reach out in [#package-ecosystem](https://getdbt.slack.com/archives/CU4MRJ7QB/) in the [dbt Community Slack](https://getdbt.com/community).
```
dbeatty10, line 9: 🙌

package-best-practices.md
```diff
@@ -0,0 +1,41 @@
...
+    - SHOULD NOT pin to a patch version of the dependent package unless they are aware of an incompatibility between their package and the dependency.
+    - SHOULD reference the version in the dbt Package Hub when available, as opposed to a `git` installation.
+
+### Interoperability
+- Packages MUST NOT override dbt Core behaviour in such a way as to impact other dbt resources (models, tests, etc) not provided by the package.
+- Packages SHOULD disambiguate their resource names to avoid clashes with nodes that are likely to already exist in a project. 
+    - For example, packages SHOULD NOT provide a model simply called `users`.
+
+### Updates
+- Packages SHOULD follow the guidance of the [Semantic Versioning Specification](https://semver.org/).
```
dbeatty10, line 41:
```suggestion
### Releases and updates
- Packages SHOULD follow the guidance of the [Semantic Versioning Specification](https://semver.org/).
- Packages MUST create a git tag with the version number that follows the accepted version format.
```
We'd probably need to be more clear about what the "accepted version format" is -- currently it is _strictly_ https://semver.org/spec/v2.0.0.html but would be relaxed slightly if #130 is merged.

package-best-practices.md
```diff
@@ -0,0 +1,41 @@
...
+Although the dbt Labs team completes a cursory review of new packages before adding them to the Package Hub, ultimately you are responsible for your package's performance.
+
+To avoid ambiguity, the bullets below are pretty formal, but we are a friendly bunch! If you need help preparing a package for deployment to the dbt Package Hub, reach out in [#package-ecosystem](https://getdbt.slack.com/archives/CU4MRJ7QB/) in the [dbt Community Slack](https://getdbt.com/community).
+
+---
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+### First run experience
+- Packages SHOULD contain a README file which explains how to get started with the package and customise its behaviour.
```
dbeatty10, line 16:
```suggestion
### First run experience
- Packages SHOULD contain a README file which explains how to get started with the package and customise its behaviour.
- The documentation SHOULD indicate which data warehouses are expected to work with this package.
```

dbeatty10 (dismissed):
So excited about this! Looks awesome 🤩 

This looks good enough to me to merge as-is -- we could always revisit with updates as-needed.

The review comments include responses and ideas.

The GitHub emojis don't have nearly a wide enough range to express how excited I am about this! 🏆

joellabes:
Shipping it! Thanks for the feedback @dbeatty10 - I've taken a very literal approach to defining version tags, and just said "it has to meet the regex". I think that's probably the best we can do if we have taken parts of two version tagging systems? If we have taken all of PEP-440 then this gets easier but I don't think we did.

package-best-practices.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Interoperability
+- Packages MUST NOT override dbt Core behaviour in such a way as to impact other dbt resources (models, tests, etc) not provided by the package.
+- Packages MUST use the cross-database macros built into dbt Core where available, such as `{{ datediff }}` and `{{ type_string() }}`.
+- Packages SHOULD disambiguate their resource names to avoid clashes with nodes that are likely to already exist in a project. 
+    - For example, packages SHOULD NOT provide a model simply called `users`.
+
+### Releases and updates
+- Packages' git tags MUST validate against the regex defined in [version.py](/hubcap/version.py).
+- Packages SHOULD follow the guidance of the [Semantic Versioning Specification](https://semver.org/).
```
dbeatty10, line 46: I don't foresee the SemVer specification changing anytime soon, but we specifically referring to 2.0.0, I think.

Optional update:
```suggestion
- Packages SHOULD follow the guidance of [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html).
```

package-best-practices.md
```diff
@@ -0,0 +1,46 @@
...
+### Dependencies
+#### Dependencies on dbt Core
+- Packages requiring a minimum version of `dbt-core` MUST declare it in the `require-dbt-version` property of `dbt_project.yml`.
+- Packages requiring a minimum version of `dbt-core` SHOULD allow all subsequent minor and patch releases of that major version. 
+    - For example, a package which depends on functionality added in dbt Core 1.2 SHOULD set its `require-dbt-version` property to `[">=1.2.0", "<2.0.0"]`.
+#### Dependencies on other packages defined in `packages.yml`:
+- Packages SHOULD import their dependencies from the dbt Package Hub when available, as opposed to a `git` installation.
+- Packages SHOULD NOT pin to a patch version of their imported package unless they are aware of an incompatibility.
+- Packages importing another package which has reached major version one or later SHOULD allow all subsequent minor and patch releases of that major version. 
+- Packages importing another package which is major version zero MAY pin to the current minor version only.
```
dbeatty10, line 36: I think the last 3 bullet points can all be replaced with simple guidance:

```suggestion
- The widest version range possible SHOULD be specified to give the dependency resolver the appropriate latitude.
```

My guess is that after lots of discussion exploring many scenarios, ☝️ will still hold as an essential summary, a [cataclysm sentence](https://radiolab.org/episodes/cataclysm-sentence) that "contains the most information in the fewest words".

The three bullet points make assumptions about the versioning scheme being used by the imported package which _might_ be strict [SemVer 2.0](https://semver.org/spec/v2.0.0.html), but just as likely might be a variant or something else entirely.

package-best-practices.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Interoperability
+- Packages MUST NOT override dbt Core behaviour in such a way as to impact other dbt resources (models, tests, etc) not provided by the package.
+- Packages MUST use the cross-database macros built into dbt Core where available, such as `{{ datediff }}` and `{{ type_string() }}`.
```
dbeatty10, line 40: `MUST` or `SHOULD`? Could a BigQuery-specific spend package BigQuery-isms? Or what if the current cross-database implementation doesn't cover some edge case that the package needs or wants to accommodate?

Optional ideas for re-wording slightly:
```suggestion
- Packages SHOULD use the [cross-database macros](https://docs.getdbt.com/reference/dbt-jinja-functions/cross-database-macros) built into dbt Core where available, such as `{{ except() }}` and `{{ type_string() }}`.
```

dbeatty10 (changes requested):
This is really **excellent** work -- looks fantastic 🤩 

I left comments for a few nit-picky things. The "wide version ranges" comment deserves the most focus and consideration.

package-best-practices.md
joellabes: >The three bullet points make assumptions about the versioning scheme being used by the imported package which might be strict [SemVer 2.0](https://semver.org/spec/v2.0.0.html), but just as likely might be a variant or something else entirely.

Could we change that `SHOULD` from Releases and Updates to a `MUST`? 

Those bullets are written in the same spirit as when you go to a fancy hotel and there's a _No Elephants Allowed In the Ballroom_ sign - you know they didn't put it up until they had to. 

The first one (no patches) is the most common and catastrophic failure state I see; I'm not wed to the others but they seemed like the obvious next question: "If I'm not allowed to pin to a patch, what _do_ you recommend?"

What do you think of something like this?
> Packages SHOULD specify the widest possible range of supported versions, to minimise issues in dependency resolution. In particular, packages SHOULD NOT pin to a patch version of their imported package unless they are aware of an incompatibility.

package-best-practices.md
joellabes: Went back and forth on this and maybe came down too hard 🤔. How about packages intended for use with multiple warehouses MUST...

I like the link to the docs

joellabes: If there's an edge case that the macros don't cover, then I think the "where available" works as an escape hatch

package-best-practices.md
```diff
@@ -0,0 +1,44 @@
...
+#### Dependencies on dbt Core
+- Packages requiring a minimum version of `dbt-core` MUST declare it in the `require-dbt-version` property of `dbt_project.yml`.
+- Packages requiring a minimum version of `dbt-core` SHOULD allow all subsequent minor and patch releases of that major version. 
+    - For example, a package which depends on functionality added in dbt Core 1.2 SHOULD set its `require-dbt-version` property to `[">=1.2.0", "<2.0.0"]`.
+#### Dependencies on other packages defined in `packages.yml`:
+- Packages SHOULD import their dependencies from the dbt Package Hub when available, as opposed to a `git` installation.
+- Packages SHOULD specify the widest possible range of supported versions, to minimise issues in dependency resolution. 
+    - In particular, packages SHOULD NOT pin to a patch version of their imported package unless they are aware of an incompatibility.
+### Interoperability
+- Packages MUST NOT override dbt Core behaviour in such a way as to impact other dbt resources (models, tests, etc) not provided by the package.
```
joellabes, line 37: @jtcohen6 I wrote this thinking about the aftermath of that Fivetran package that overrode freshness collection and had unfortunate side effects. 

But, if someone like [this community member](https://getdbt.slack.com/archives/C01NH3F2E05/p1659364136316949) turned their [gist](https://gist.github.com/LewisDavies/f126271284df804c3cf5cc9b094db6a8) into a package, that would be an acceptable overriding of behaviour yeah? Does this rule need a bit of wordsmithing or do we just drop it down to `REALLY SHOULD NOT` per [RFC 6919](https://datatracker.ietf.org/doc/html/rfc6919)?

package-best-practices.md
dbeatty10: > Packages SHOULD specify the widest possible range of supported versions, to minimise issues in dependency resolution. In particular, packages SHOULD NOT pin to a patch version of their imported package unless they are aware of an incompatibility.

🎯 looks perfect

> Could we change that `SHOULD` from Releases and Updates to a `MUST`?

I'm _guessing_ that we can't say "Packages MUST follow the guidance of Semantic Versioning 2.0.0" because we don't have any auditing mechanism -- there is no easy way to validate to what degree a package is following it or not. We _can_ validate the version string with a regex, but we _can't_ reasonably determine if they are following the SemVer guidance or not. They are on the honor system for this one.

package-best-practices.md
dbeatty10: Love your suggestion to introduce the nuanced language of [RFC 6919](https://datatracker.ietf.org/doc/html/rfc6919) where it is useful! 🤩 

News flash: I'm not an expert on the interpretation and application of RFC 2119 or 6919. The following is all informed by my nascent understanding and interpretations.

After reading the ["Guidance in the use of these Imperatives"](https://datatracker.ietf.org/doc/html/rfc2119#section-6) section of RFC 2119, it feels like `SHOULD` is the applicable imperative here. The following stood out to me in particular:
> they must not be used to try to impose a particular method on implementors where the method is not required for interoperability.

A package could ignore some of our "musts" and operate just fine. I didn't check, but I'm guessing there are examples currently in the [Hub](https://hub.getdbt.com/) that would back this up.

But most (not all) of our current usage of `MUST` seems like it should be `SHOULD` instead. `SHOULD` appears to be a **really** strong word in the spec! It shouldn't be confused at all with `OPTIONAL`.

So I think we shouldn't feel like we are pandering or compromising on our expectations in any way when we use `SHOULD`. We should feel like we're laying down something strong, confident, and very assertive.

package-best-practices.md
dbeatty10: +1 for `REALLY SHOULD NOT`

dbeatty10:
Re-reading the SemVer 2.0.0 spec and specifically the [FAQ](https://semver.org/spec/v2.0.0.html#how-do-i-know-when-to-release-100), we should strongly consider adding the following:
> Package versions SHOULD be a minimum of 1.0.0 to align with the spirit of [semantic versioning for production software](https://semver.org/spec/v2.0.0.html#how-do-i-know-when-to-release-100). If a package isn't ready for production, then it isn't yet ready for the Hub.

If the Hub is intended to also host packages that _should not_ be used by anyone in production, then I fully retract the suggestion above.

package-best-practices.md
joellabes: Tragically, 6919 was published on **1 April** 2013 and was never ratified. (See also https://datatracker.ietf.org/doc/html/rfc2324)

Happy to roll with SHOULD here

package-best-practices.md
joellabes: Fair!

package-best-practices.md
joellabes: `REALLY SHOULD NOT` doesn't actually exist 😢 (discussed above) - should've made clearer that this was a joke

joellabes:
> Re-reading the SemVer 2.0.0 spec and specifically the [FAQ](https://semver.org/spec/v2.0.0.html#how-do-i-know-when-to-release-100), we should strongly consider adding the following:
> 
> > Package versions SHOULD be a minimum of 1.0.0 to align with the spirit of [semantic versioning for production software](https://semver.org/spec/v2.0.0.html#how-do-i-know-when-to-release-100). If a package isn't ready for production, then it isn't yet ready for the Hub.
> 
> If the Hub is intended to also host packages that _should not_ be used by anyone in production, then I fully retract the suggestion above.

Yeahhh... but.... we are nowhere near following that rule ourselves, nor are any of the other packages in the Hub. The rest of this document is is codifying existing behaviour to save us from abstractly saying "you don't meet our norms". This would be a change of policy which I don't really want to spend time enforcing at the moment.

And FWIW, we already say that people SHOULD follow SemVer, so I don't think we need to call out one FAQ specifically. I'd be happy to suggest to people who are submitting future packages that they might want to tag them 1.0.0 instead of 0.1.0 though

dbeatty10:
I'm attracted to encoding this for a couple reasons:
- Many projects that _look_ like SemVer and _claim to be_ SemVer are really just [ZeroVer](https://0ver.org/)
- No one will be harmed by diving in at 1.0.0 -- c'mon in! the water's fine! 🏄 
- Acceptable version numbers themselves are one of the very few things we can actually programmatically enforce for new packages.

I'm very aware of our own failure to follow that one 🤒 Probably even more reason for us to turn the tide in ways that we can.

joellabes:
What about things like [dbt-project-evaluator](https://github.com/dbt-labs/dbt-project-evaluator), [dbt-audit-helper](https://github.com/dbt-labs/dbt-audit-helper) or [dbt-codegen](https://github.com/dbt-labs/dbt-codegen)? 

They're primarily for interactive use and I wouldn't necessarily encourage people to use in an _unmonitored_ production environment, especially not codegen. 

I am particularly uncomfortable with the editorialising of "if it's not ready for prod it's not ready for the Hub". Can we meet in the middle and put an addendum on the "should follow SemVer" along the lines of "Note in particular the recommendation for production-ready packages to be version 1.0.0 or above"?

dbeatty10:
### Production-ready packages
> Note in particular the recommendation for production-ready packages to be version 1.0.0 or above

This would be movement in the right direction.

Would it end up looking something like this?

```markdown
### Releases and updates
- Packages' git tags MUST validate against the regex defined in [version.py](/hubcap/version.py).
- Packages SHOULD follow the guidance of [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html).
    - Note in particular the recommendation for production-ready packages to be version `1.0.0` or above
```

### Right-shifted SemVer

> What about things like [dbt-project-evaluator](https://github.com/dbt-labs/dbt-project-evaluator), [dbt-audit-helper](https://github.com/dbt-labs/dbt-audit-helper) or [dbt-codegen](https://github.com/dbt-labs/dbt-codegen)?

What constitutes being "prod"? 🤷

To the best that I can tell, some or all of those projects follow "right-shifted SemVer" -- just shift everything one separator to the left to recover the `MAJOR.MINOR.PATCH` semantic meaning! (`PATCH` will always be `0` following this shift.)

If that is the case, then there's "major" releases hiding in there along with an implied public API -- they could have just started out at `1.0.0` and then it's less complicated to understand the semantic meaning of each of their versions.

Starting out using [ZeroVer](https://0ver.org/) probably has measurable gravity that is hard to escape.

Interestingly, the SemVer FAQ explains scenarios when the bump to 1.0.0 is later than it should be. It doesn't explain scenarios when it is _too_ early though.

joellabes:
yeah I think you’re probably right that it's mostly just right shifted semver. im having horrible flashbacks to high school calculus when we learned about integration and always had to remember the `+ c` to reflect whatever constant was lost and unrecoverable. 

I feel good about this now - we're encouraging good behaviour for new packages, we can gently nudge the existing ecosystem to make better choices over time, but we're not marking the entire package hub as non-compliant which feels a bit rough. I also suspect that this is another place where we were following in dbt Core's footsteps: if _it_ had started at v1, would all the packages be zero-based? I doubt it.

package-best-practices.md
```diff
@@ -0,0 +1,46 @@
...
+---
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+### First run experience
+- Packages SHOULD contain a README file which explains how to get started with the package and customise its behaviour.
+- The documentation SHOULD indicate which data warehouses are expected to work with this package.
+
+### Customisability 
+- Packages MUST NOT hard-code table references, and MUST use `ref` or `source` instead.
```
dbeatty10, line 20: Non-blocking commentary.

The current text definitely conveys the _spirit_ of the guidance, but an overly lawyerly maintainer could hard-code a view reference and claim full compliance.

Two opportunities for further clarity:
- table → relation
- use applicable adapter methods rather than directly utilizing `describe`, `show`, or `information_schema`

I'm fuzzy on the detailed guidance to accomplish the latter, so that's why I don't have an explicit suggestion that covers both opportunities.

Our glossary doesn't include an entry "relation" currently, it only has the following disparate entries:
- https://docs.getdbt.com/terms/table
- https://docs.getdbt.com/terms/view
- https://docs.getdbt.com/terms/cte

The closest documentation I could find quickly focuses on the definition and usage of the Relation Python object:
- https://docs.getdbt.com/reference/dbt-classes#relation

dbeatty10 (approved):
👍 This will add great value.

Approving with the belief and trust that we'll make edits in the future as-needed.

PR #145, opened 2022-08-11
Add upstream-prod package

LewisDavies:
Adding my new package to the package hub

https://github.com/LewisDavies/upstream-prod

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -252,11 +252,13 @@
     "everpeace": [
         "dbt-models-metadata"
     ],
-
     "infinitelambda": [
         "carbon-analytics"
     ],
     "andrewcstewart": [
         "dbt-uaparser"
+    ],
+    "LewisDavies": [
+        "upstream-prod"
     ]
 }


dbeatty10:
@LewisDavies cool! Overriding `ref` is kinda fun and magical ✨  -- when I [first read about it](https://docs.getdbt.com/reference/dbt-jinja-functions/builtins#usage), I was like 🤯!

Everything looks good to me, but I want to get Joel's confirmation on one piece:
- @joellabes, overriding built-ins is totally fine, right? i.e., the [prohibition of overriding dbt Core behavior](https://github.com/dbt-labs/hubcap/blob/main/package-best-practices.md#interoperability) is referring to something totally different, right?

LewisDavies:
Cheers @dbeatty10! Just to clarify, the package won't directly override `ref` unless users explicitly change their models. After installing, you would need to change every instance `{{ ref() }}` to `{{ upstream_prod.ref() }}` for it to work.

I tried to find a way for users to optionally override the default macro, but based on [this issue](https://github.com/dbt-labs/docs.getdbt.com/issues/971) it doesn't seem possible right now.

LewisDavies:
Ah, gotcha! I like the idea of using an overrides directory.

Out of curiosity, would it be OK to add those commands to the package as a shell script?

dbeatty10:
@LewisDavies After discussion with colleagues, here's a better suggestion that we feel good about:
- provide instructions how consumers can add a thin wrapper around the package macro

You'd just tell consumers how to also add this file to their dbt project at the same time they add your package to their packages.yml.

Here's roughly what the thin wrapper would look like:
`macros/ref.sql`
```sql
{% macro ref(model_name) %}

  {% do return(upstream_prod.ref(model_name)) %}

{% endmacro %}
```

You'd probably need to add parameters within the wrapper to make it work the way you want with `upstream_prod`.

LewisDavies:
Nice, it's great that it won't need to change unless I add new parameters. I'll get that in the docs soon, thanks!

joellabes (approved):
Thanks @LewisDavies!

Issue #95, opened 2022-02-10
Register a dbt package from a sub directory of a repository 


turbo1912:
Hi friends, 

We are about to release a new dbt package and we wanted to keep everything as a morepo in our project. That would mean instead of pointing to a github repository we would have to point it to a subdirectory in our existing repository for example; https://github.com/fal-ai/fal/feature-store. 

I don't think this is possible right now, but we are happy to contribute if this is something you would like to see implemented. 

I just had a quick glance at the code and looks like a change like this is self contained in the hubcap repository and I don't have to touch the logic how `dbt-core` downloads dependencies.

elongl:
We'd really love this feature as well!
Will also like to contribute if possible.

joellabes:
@turbo1912 and @elongl thanks for opening this, and sorry that it took a while to get back to you! 

Conceptually, I'm very supportive of this, especially if it doesn't impact how dbt-core downloads things. 

The things that are giving me pause are that we don't have any testing around the project (other than a couple of semver tests that @dbeatty10 added in https://github.com/dbt-labs/hubcap/pull/133) at the moment so can't really guarantee that changes don't cause downstream issues. 

I would say go for it, with the disclaimer that code review might be slow as we try to rustle up someone who knows how hubcap works. We'll also probably wind up dragging in someone from @dbt-labs/core to double check that there aren't any flow-on effects.

elongl:
Awesome! Glad to hear you and the team are on board.
I'll send here in a couple of days a rough design of how I'm planning to implement it so I could get your approval and start working on it. Thanks a lot.

jtcohen6:
I definitely see the value here!

I think this may be tricky for us to do in the current implementation of Hubcap + Hub site (hub.getdbt.com), though not impossible. The Hub site does not actually store/mirror specific files—it's just a pointer to a GitHub tarball URL, containing a zipped version of _all_ files from the repo. E.g. for `dbt_utils` version 0.8.6: https://codeload.github.com/dbt-labs/dbt-utils/tar.gz/0.8.6

If we wanted to proceed with this as an extension of the current implementation, I think it would need to include:
- A new Hub API field, `subdirectory`
[...]

PR #146, opened 2022-08-12
Add `require_dbt_version` to Hub API

jtcohen6:
resolves https://github.com/dbt-labs/hub.getdbt.com/issues/165

~What is `works_with` used for? Is this the old name for `packages`? It's not being used for anything by recent versions of `dbt-core`.~

I've added a totally new field to the hub API (`require_dbt_version`). ~I just started by reusing the existing field for expediency.~

I tested the simple methods locally, against `dbt_project.yml` files that do and don't have `require-dbt-version` defined. Note that this _will not_ work unless `require-dbt-version` is a valid semver string/list (as documented), so we probably want to add some validation logic to that effect.

As far as moving forward with this:
- Not sure of the right way to end-to-end test this
- Would we want to rerun the Hubcap script, to update existing package versions?

### DIFF
--- a/hubcap/records.py
+++ b/hubcap/records.py
@@ -73,12 +73,13 @@ class UpdateTask(object):
             os.chdir(self.local_path_to_repo)
             git_helper.run_cmd(f'git checkout tags/{tag}')
             packages = package.parse_pkgs(Path(os.getcwd()))
+            require_dbt_version = package.parse_require_dbt_version(Path(os.getcwd()))
 
             # return to hub and build spec
             os.chdir(main_dir)
             package_spec = self.make_spec(
                 self.github_username, self.github_repo_name,
-                self.package_name, packages, tag
+                self.package_name, packages, require_dbt_version, tag
             )
 
             version_path = self.hub_version_index_path / Path(f'{tag}.json')
@@ -164,7 +165,7 @@ class UpdateTask(object):
         print("      SHA1: {}".format(digest))
         return digest
 
-    def make_spec(self, org, repo, package_name, packages, version):
+    def make_spec(self, org, repo, package_name, packages, require_dbt_version, version):
         '''The hub needs these specs for packages to be discoverable by deps and on the web'''
         tarball_url = "https://codeload.github.com/{}/{}/tar.gz/{}".format(org, repo, version)
         sha1 = self.get_sha1(tarball_url)
@@ -176,7 +177,7 @@ class UpdateTask(object):
             "version": version,
             "published_at": "1970-01-01T00:00:00.000000+00:00",
             "packages": packages,
-            "works_with": [],
+            "works_with": require_dbt_version,
             "_source": {
                 "type": "github",
                 "url": "https://github.com/{}/{}/tree/{}/".format(org, repo, version),
--- a/hubcap/records.py
+++ b/hubcap/records.py
@@ -177,7 +177,8 @@ class UpdateTask(object):
             "version": version,
             "published_at": "1970-01-01T00:00:00.000000+00:00",
             "packages": packages,
-            "works_with": require_dbt_version,
+            "require_dbt_version": require_dbt_version,
+            "works_with": [],
             "_source": {
                 "type": "github",
                 "url": "https://github.com/{}/{}/tree/{}/".format(org, repo, version),
--- a/hubcap/package.py
+++ b/hubcap/package.py
@@ -31,6 +31,12 @@ def parse_pkg_name(repo_dir):
         return name if name else ''
 
 
+def parse_require_dbt_version(repo_dir):
+    with open(repo_dir / Path('dbt_project.yml'), 'r') as stream:
+        requirements = yaml.safe_load(stream).get('require-dbt-version', [])
+        return requirements if requirements else []
+
+
 def parse_pkgs(repo_dir):
     if os.path.exists(repo_dir / 'packages.yml'):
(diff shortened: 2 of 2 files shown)


hubcap/records.py
```diff
@@ -176,7 +177,7 @@ def make_spec(self, org, repo, package_name, packages, version):
             "version": version,
             "published_at": "1970-01-01T00:00:00.000000+00:00",
             "packages": packages,
-            "works_with": [],
+            "works_with": require_dbt_version,
```
jtcohen6, line 180: `require-dbt-version` can be a string or a list ([docs](https://docs.getdbt.com/reference/project-configs/require-dbt-version)), so the resulting value in the Hub API can be a string or a list.

I added [logic to handle this](https://github.com/dbt-labs/dbt-core/pull/5651/files#diff-56648fa17b1711068e02ac1031a915b3e2dbc667a973b057c2c711e69054c134R136-R139) within `dbt-core`, rather than having that logic live here (= turning `string` into `[string]`).

joellabes:
I'd be keen to call it `require_dbt_version` - works with could be many things. dbt Core version and specific [adapters](https://github.com/dbt-labs/dbt-core/issues/1471) spring to mind, but also more left field things like "does it work with Core and Cloud", or "does it work without superuser access to the database"...

joellabes:
> * Would we want to rerun the Hubcap script, to update existing package versions?

I think so, yeah!

joellabes (approved):
This looks good!

> Note that this _will not_ work unless `require-dbt-version` is a valid semver string/list (as documented), so we probably want to add some validation logic to that effect.

I don't see this anywhere; what are the consequences of an invalid string? Does dbt throw an exception if required version is something non semver, or does it ignore it? 

I assume it ignores it and runs as normal? Which is ok to me, the only part that offend me on a principled level is that we might put inaccurate data into a long lived json file.

jtcohen6:
> Does dbt throw an exception if required version is something non semver, or does it ignore it?

Throwing an exception is the current behavior in https://github.com/dbt-labs/dbt-core/pull/5651. So if it finds `require-dbt-version: ["1.2.3.4"]`:
```
dbt.exceptions.SemverException: "1.2.3.4" is not a valid semantic version.
```
But I could definitely switch it to just start ignoring `require-dbt-version` if it's invalid. Same as if it's missing / empty.

FWIW, that's the same exception the package author would see any time they try running that package/project with an invalid `require-dbt-version` config. So I don't think it's very likely for an invalid one to slip in accidentally. I just bring it up because we're not running dbt or doing any dbt-related validation within the Hubcap process.

Once this PR is merged, I'm not sure of the right way to go about updating / re-deploying all existing package versions.

joellabes:
Exceptions at development time are even better! The reason I was at peace with it was mostly practical: we don't have a practical way to auto-open an issue against a misbehaving repo to warn them, and I didn't want the dbt team to keep getting pings about packages that we have no control over. 

> Once this PR is merged, I'm not sure of the right way to go about updating / re-deploying all existing package versions.

Is it not just this bit everywhere? https://github.com/dbt-labs/hub.getdbt.com/pull/1770 ?

jtcohen6:
> Is it not just this but everywhere? https://github.com/dbt-labs/hub.getdbt.com/pull/1770 ?

You're right, it is just that but everywhere :)

I took some shortcuts / creative liberties to prepare that PR, while running Hubcap locally. Could do the same thing but more.

dbeatty10:
> Is it not just this but everywhere? [dbt-labs/hub.getdbt.com#1770](https://github.com/dbt-labs/hub.getdbt.com/pull/1770) ?

Yep, I think you are exactly right about [dbt-labs/hub.getdbt.com#1770](https://github.com/dbt-labs/hub.getdbt.com/pull/1770) being the exemplar, @joellabes.

When a pull request like that is merged in the `hub.getdbt.com` repo, my understanding is that it immediately triggers a re-deploy on Netflify somehow. I haven't personally seen how Netflify is configured. But I've observed that it immediately redeploys when pull requests generated by the hubcap.py process are auto-merged.

dbeatty10:
> Could do the same thing but more.

Yes, more (locally modified) cowbell should do the trick here. 🐮 

Side note: I suspect the same general strategy will be needed if [dbt-labs/hubcap #95](https://github.com/dbt-labs/hubcap/issues/95) goes forward.

dbeatty10 (approved):
Thanks for diving into this across like three different repos @jtcohen6 !

To connect the dots for those interested, here are two other related pull requests:
1. [dbt-labs/dbt-core #5651](https://github.com/dbt-labs/dbt-core/pull/5651)
1. [dbt-labs/hub.getdbt.com #1770](https://github.com/dbt-labs/hub.getdbt.com/pull/1770)

This particular PR will generate JSON blobs like this (`require_dbt_version` being the new key+values):
```json
    "id": "dbt-labs/dbt_utils/0.8.6",
    "name": "dbt_utils",
    "version": "0.8.6",
    "published_at": "1970-01-01T00:00:00.000000+00:00",
    "packages": [],
    "require_dbt_version": [
        ">=1.0.0",
        "<2.0.0"
    ],
    "works_with": [],
    "_source": {
        "type": "github",
```

jtcohen6:
I'm going to merge this one first. It won't be breaking for any existing versions of `dbt-core`, and it doesn't modify any existing Hub packages, but it will capture new ones that aren't included in the "manual backfill" I ran last night (https://github.com/dbt-labs/hub.getdbt.com/pull/1779).

PR #150, opened 2022-10-19
Added ScalefreeCOM/datavault4dbt

tkirschke:
I have added the datavault4dbt package from ScalefreeCOM.

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -26,6 +26,9 @@
     "Snowflake-Labs": [
         "dbt_constraints"
     ],
+    "ScalefreeCOM": [
+        "datavault4dbt"
+    ],
     "Teradata": [
         "dbt-teradata-utils"
     ],


joellabes (approved):
Welcome to the Hub! 

While reviewing against our [best practices](https://github.com/dbt-labs/hubcap/blob/main/package-best-practices.md), I noticed that your dbt utils version range (`[">=0.8.0", "<0.9.0"]`) is probably unnecessarily narrow. It excludes version 0.9.x which is the current shipping version.

tkirschke:
> Welcome to the Hub!
> 
> While reviewing against our [best practices](https://github.com/dbt-labs/hubcap/blob/main/package-best-practices.md), I noticed that your dbt utils version range (`[">=0.8.0", "<0.9.0"]`) is probably unnecessarily narrow. It excludes version 0.9.x which is the current shipping version.

Hi Joel and thanks for your feedback!
We have now raised the dbt_utils requirements!

PR #151, opened 2022-10-19
Add dbt-snowflake-monitoring package to dbt hub

NiallRees:
https://github.com/get-select/dbt-snowflake-monitoring


dbeatty10 (approved):
Very useful @NiallRees and @ian-whitestone 🤩

Also appreciate that you are adding to the hub starting at 1.0.0 -- I think everyone should do that and not anything lower!

PR #154, opened 2022-10-31
Enable pull requests to be either consolidated or individual

dbeatty10:
## Problem
Currently, [github.com/dbt-labs/hub.getdbt.com](https://github.com/dbt-labs/hub.getdbt.com/pulls) is limited to merging only a single pull request per hour. That means a maximum of 24 new versions can be released per day. As an unfortunate side-effect, tons of duplicate pull requests also get generated.

## Solution

This pull request achieves multiple things:
1. commits can _optionally_ be gathered into a **single pull request** rather than one pull request per repo
1. an alternative destination repo for pull requests can be configured
    - this is handy for testing & development
1. the `git-tmp` directory can be preserved across executions
    - this is handy for inspecting the state of branches and commits

## Why

We ❤️ our friends at Fivetran (e.g, @fivetran-joemarkiewicz @fivetran-sheringuyen, etc), and switching on the **single/consolidated pull request** option would allow them to release with impunity.

## On the other hand
The **individual pull requests** option is the current behavior, and intend for it to still be the default immediately after this PR is merged. Hopefully we can switch to the consolidated PR option some time afterwards, and it will be easy to switch back to the original if we want.

### consolidated PR
To switch to a single consolidated PR instead of individual PRs:
- update the `CONFIG` environment variable to contain `"one_branch_per_repo": false` rather than `"one_branch_per_repo": true`.

### development target repo
To use an alternative (development) repo rather than the production repo:
- update the `CONFIG` environment variable to contain something like `"repo": "hub.getdbt.com-test"` rather than `"repo": "hub.getdbt.com"`.

This assumes that a target repo with the name `hub.getdbt.com-test` has already been created and published!

### preserve clones across executions
To preserve the `git-tmp` directory between (local) executions:
- Execute with `./cron.sh --no-cleanup` rather than just `./cron.sh`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,10 +4,11 @@
 
 ### Personal access token (PAT)
 
-Follow [these](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) instructions to configurate a personal access token.
+Follow [these](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) instructions to configure a PAT.
 
-Only permissions needed:
-- public_repo
+Scopes:
+- repo (and all sub-items)
+- workflow
 
 Save the token to a secure location. Click the "Configure SSO" button and "Authorize" if applicable.
 
@@ -49,7 +50,7 @@ export ENV=test
 ```
 
 ### Optional configuration environment variables
-```
+```shell
 # Default value is the `git-tmp` directory within the current working directory
 # This directory will be deleted by default at the end of the run
 export GIT_TMP=git-tmp
@@ -57,7 +58,7 @@ export GIT_TMP=git-tmp
 
 ### Optional parameters
 Preserve commits/build artifacts within the `$GIT_TMP` directory
-```
+```shell
 export ENV=test
 ./cron.sh --no-cleanup
 ```
--- a/config.example.json
+++ b/config.example.json
@@ -3,7 +3,8 @@
         "name": "github_user",
         "token": "pe4s0n@l-@cce$$-t0k3n"
     },
-    "remote": "git@github.com:dbt-labs/hub.getdbt.com.git",
+    "org": "dbt-labs",
+    "repo": "hub.getdbt.com",
     "push_branches": true,
     "one_branch_per_repo": true
 }
--- a/cron.sh
+++ b/cron.sh
@@ -75,7 +75,7 @@ if [ "$ENV" = 'prod' ] || [ "$ENV" = 'test' ]; then
     PRIOR_GIT_NAME="$(git config --global user.name || : )"
 
     # Setup git repo for automated commits during execution
-    git config --global user.email 'drew@fishtownanalytics.com'  # TODO: make this a dedicated CI user
+    git config --global user.email 'buildbot@fishtownanalytics.com'
     git config --global user.name 'dbt-hubcap'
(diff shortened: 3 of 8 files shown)


joellabes:
I see that there are now two strategies which can be configured, which makes sense. It just looks like too little code has changed! Is it basically just a case of changing the name of the PR so that there isn't a new one for every package, and checking whether there's already a PR in flight that can have new code attached to it? 

TL;DR: I am extremely into the outcome of this! it's not 100% clear to me _how_ it gets there, but I'm sorta willing to take it on faith 📿

dbeatty10:
Branch name for [current behavior](https://github.com/dbt-labs/hubcap/pull/154/files#diff-513d3c12aa63bb0675ce76834c791fd04a639e99c111d8d6439e161fa1ebc621R34):
```python
f"bump-{org}-{repo}-{setup.NOW}"
```

Branch name for [new option](https://github.com/dbt-labs/hubcap/pull/154/files#diff-513d3c12aa63bb0675ce76834c791fd04a639e99c111d8d6439e161fa1ebc621R42):
```python
f"bump-package-versions-{setup.NOW}"
```

The first gives a unique branch name per Hub package, and the latter uses the same branch name across all Hub packages. (`setup.NOW` is just a unique timestamp per script invocation, so it is a constant.)

As a final step, hupcap creates one pull request per branch. 👈 This code already existed, so that's why so little changed to enable the new behavior.

joellabes:
@dbeatty10 OK so it's sorta like how at the moment hubcap will happily create a single PR for multiple releases of a single package (e.g. metrics 0.1.2 and 0.1.3 if they're both cut within the same hour)? Makes sense 🎉

dbeatty10:
> @dbeatty10 OK so it's sorta like how at the moment hubcap will happily create a single PR for multiple releases of a single package (e.g. metrics 0.1.2 and 0.1.3 if they're both cut within the same hour)? Makes sense 🎉

Yep, you got it 🎉

fivetran-joemarkiewicz:
@dbeatty10 and @joellabes this is awesome! Thanks so much for taking our packages into consideration and adding this functionality. 

Looking ahead to our dbt-utils v1.0.0 migrations for **all** +80 Fivetran dbt packages, is there anything on our end we should add or need to consider before cutting releases in order to take advantage of this feature?

dbeatty10:
@fivetran-joemarkiewicz There will be nothing you need to add on your end.

Although this feature has been released, it's hidden behind a feature toggle and we haven't turned this feature flag on quite yet. I'll let you know when we do!

From there, you can let me know when you want to "release the hounds", and we can do a test with a small handful to confirm it actually works as intended.

fivetran-joemarkiewicz:
Sounds great! I will be sure to let you know once the hounds 🐕 🐩 🦤 are ready and we can coordinate from there.

dbeatty10:
@fivetran-joemarkiewicz and @fivetran-sheringuyen:

Just wanted to let you know that hubcap should be ready for you to bump versions across multiple Fivetran packages simultaneously. Here's a couple examples in action:
- https://github.com/dbt-labs/hub.getdbt.com/pull/2077
- https://github.com/dbt-labs/hub.getdbt.com/pull/2074

The script still only runs once per hour, but if you tag new versions for 25 different Fivetran packages in GitHub between 4 and 5 PM, then all 25 should have their versions bumped on hub.getdbt.com by 5:15 PM.

PR #160, opened 2022-11-01
Use project git config instead of global

### LINKED ISSUE #153: Use project git config instead of global
`git` config is updated globally [here](https://github.com/dbt-labs/hubcap/blob/35b451b5fe2b052302f2c0baa4e0ba0426aa7036/cron.sh#L72-L80) in a bash script. However, we can use something like [this](url) using Python instead. It can be placed after the clone [here](https://github.com/dbt-labs/hubcap/blob/35b451b5fe2b052302f2c0baa4e0ba0426aa7036/hubcap/hubcap.py#L28-L29).

## Advantages
- Scope the config specific to where it is needed rather than globally
- No need to clean-up afterwards if it fails in the middle of execution
- The code should be shorter overall
- More people know how to read and write Python than bash
- One step closer to removing the bash script altogether

## Implementation details

The `config.example.json` should be updated to something like the following:
```json
{
    "user": {
        "name": "dbt-hubcap",
        "email": "buildbot@fishtownanalytics.com",
        "token": "pe4s0n@l-@cce$$-t0k3n"
    },
    "org": "dbt-labs",
    "repo": "hub.getdbt.com",
    "push_branches": true,
    "one_branch_per_repo": true
}
```

Then the `name` and `email` keys should be used to set the project config for the git repo.

dbeatty10:
Resolves #153

### DIFF
--- a/hubcap/hubcap.py
+++ b/hubcap/hubcap.py
@@ -29,9 +29,10 @@ one_branch_per_repo = config.get("one_branch_per_repo", True)
 REMOTE = f"https://github.com/{github_org}/{github_repo}.git"
 PULL_REQUEST_URL = f"https://api.github.com/repos/{github_org}/{github_repo}/pulls"
 TMP_DIR = os.environ['GIT_TMP']
-GITHUB_USERNAME = config.get('user', {}).get('name', 'dbt-hubcap')
-GITHUB_EMAIL = config.get('user', {}).get('email', 'buildbot@fishtownanalytics.com')
-TOKEN = config.get('user', {}).get('token')
+GITHUB_USER = config.get('user', {})
+GITHUB_USERNAME = GITHUB_USER.get('name', 'dbt-hubcap')
+GITHUB_EMAIL = GITHUB_USER.get('email', 'buildbot@fishtownanalytics.com')
+TOKEN = GITHUB_USER.get('token')
 user_creds = {'name': GITHUB_USERNAME, 'token': TOKEN}
 PACKAGE_MAINTAINERS = setup.load_package_maintainers()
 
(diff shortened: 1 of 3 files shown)


hubcap/hubcap.py
```diff
@@ -29,7 +29,10 @@
 REMOTE = f"https://github.com/{github_org}/{github_repo}.git"
 PULL_REQUEST_URL = f"https://api.github.com/repos/{github_org}/{github_repo}/pulls"
 TMP_DIR = os.environ['GIT_TMP']
-TOKEN = config['user']['token']
+GITHUB_USERNAME = config.get('user', {}).get('name', 'dbt-hubcap')
+GITHUB_EMAIL = config.get('user', {}).get('email', 'buildbot@fishtownanalytics.com')
+TOKEN = config.get('user', {}).get('token')
```
joellabes, line 34:
```suggestion
GITHUB_USER = config.get('user', {})
GITHUB_USERNAME = config.get('user', {}).get('name', 'dbt-hubcap')
GITHUB_EMAIL = config.get('user', {}).get('email', 'buildbot@fishtownanalytics.com')
TOKEN = config.get('user', {}).get('token')
```

joellabes, line 34: A bit DRYer:

```suggestion
GITHUB_USER = config.get('user', {})
GITHUB_USERNAME = GITHUB_USER.get('name', 'dbt-hubcap')
GITHUB_EMAIL = GITHUB_USER.get('email', 'buildbot@fishtownanalytics.com')
TOKEN = GITHUB_USER.get('token')
```

joellabes (approved):
Nice! Again not familiar enough with what I assume is the GH sdk to know if the specifics are right, but I like the description of the bug you're fixing 🎉 

One comment on slightly WET code but otherwise lgtm!

hubcap/hubcap.py
dbeatty10: Yeah! 🙌

joellabes (approved):

PR #208, opened 2022-12-05
update hub.json file with saras daton packages

saras-daton:
We, at Saras Analytics, have developed 3 packages and have added them to the hub.json file.

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -23,6 +23,11 @@
     "Montreal-Analytics": [
         "dbt-snowflake-utils"
     ],
+    "Saras-Daton":[
+        "currency_exchange_rates",
+        "amazon_ads_bigquery",
+        "amazon_sellerpartner_bigquery"
+    ],
     "Snowflake-Labs": [
         "dbt_constraints"
     ],


dbeatty10:
Thanks for submitting these three packages @saras-daton !

Either myself or @joellabes should be able to give this a review within the next few days.

As part of reviewing this PR, we will provide a cursory check against [our package best practices](https://github.com/dbt-labs/hubcap/blob/main/package-best-practices.md) for these three repos:
- https://github.com/saras-daton/currency_exchange_rates
- https://github.com/saras-daton/amazon_ads_bigquery
- https://github.com/saras-daton/amazon_sellerpartner_bigquery

joellabes (changes requested):
Hey @saras-daton, thanks for submitting these!

A couple of comments: 
- I'd suggest you **not include the adapter your package is built for in its name** (i.e. the `_bigquery` suffix) - you can support multiple warehouses from a single package, and if you ever add support for others in the future then you will run into a lot of trouble migrating people over to the new name. 
  - Note that the package hub references the name in dbt_project.yml, not the name of the repository; if you're only changing one, it needs to be that one. (You can of course also change the repo name). You'll also need to do a second release of the package for the Hub to identify the changed name. 
  - When you do the second release, [consider starting from 1.0.0](https://semver.org/spec/v2.0.0.html#how-do-i-know-when-to-release-100) if your package is ready for production use cases.
- I understand that you can't reference packages that don't exist yet, but once this is merged I recommend you change the [git references](https://github.com/saras-daton/amazon_ads_bigquery/blob/main/packages.yml) to the currency_exchange_rates packages to refer to the package hub
- At a quick glance, I couldn't find any reason that the two Amazon packages needed dbt Core v1.3+ - unless there's something i missed, you could consider dropping down to v1.0 and above to maximise the number of potential users. 
- One other side note: dbt is always lowercase 😉 

These are not blockers to releasing the packages; the only reason I haven't merged it is the name of the packages is best to sort out before an initial release. If you don't plan to support other warehouses in the future then let me know and we can merge this as-is.

joellabes:
> 1. Cross Database Adapter - We have also initially planned for the same. However, our codes are very specific and requires a multi level of transformations using arrays. The codes varies due to that. Hence we would like to go ahead with separate packages.

Just to be totally clear, it's possible to have multiple variants which solve the problems in different ways but which all return the same shape table at the end. 

For example you'd do something like this: 

```sql
{% macro transform_array(arg1, arg2) %}
  {{ return(adapter.dispatch('transform_data', 'your_package_name')(arg1, arg2)) }}
{% endmacro %}

{% macro bigquery__transform_array(arg1, arg2) %}
  --BQ-specific logic here
{% endmacro %}

{% macro snowflake__transform_array(arg1, arg2) %}
  --snowflake-specific logic heere
{% endmacro %}

{% macro default__transform_array(arg1, arg2) %}
  --either a fallback ANSI SQL approach, or throw an exception for unsupported adapters
{% endmacro %}
```

and then you'd call it in your model like this 

```sql
select 
  col1, 
  col2,
  {{ your_package_name.transform_array('array_column', 'second_argument') }}
from {{ ref('some_model') }}
```

or if it's not able to be broken into small components like that, then you could even make a model that is a one-line call into `{{ package_name.do_the_whole_thing() }}` with warehouse specific implementations (I don't recommend this). 

For examples of projects doing complex JSON/array parsing in a cross-database friendly way, have a look at https://github.com/brooklyn-data/dbt_artifacts (e.g. [this file](https://github.com/brooklyn-data/dbt_artifacts/blob/main/macros/upload_test_executions.sql)) and https://github.com/dbt-labs/dbt-project-evaluator (e.g. [this file](https://github.com/dbt-labs/dbt-project-evaluator/blob/main/macros/select_from_values.sql))

saras-daton:
Hi @joellabes 

Thank you for the quick reply and the detailed explanation. Let me discuss with my team on this and will get back to you in the next couple of days.

dbeatty10:
@saras-daton dbt Labs employees will be taking a much needed holiday break until January 3rd, so we'll pick up the discussion with you again when we return. Feel free to respond any time either before then or afterwards -- we just didn't want you to feel ignored if you didn't hear back from us for a long while.

saras-daton:
Hi @dbeatty10 and @joellabes - Thanks for all suggestions and guiding us to make the code package better. We have incorporated cross database macros and made the code more generic. A new PR has been raised for the same - **"saras daton dbt hub package for amazon seller partner #244"** Request you to review and help us with the next steps.

joellabes:
Resolvd by #244

PR #210, opened 2022-12-07
Create pull_request_template.md

joellabes:
Bringing the checklist into the PR flow to help people see it!

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,32 @@
+## Description 
+
+Tell us about your new package! 
+
+## Checklist
+This checklist is a cut down version of the [best practices](package-best-practices.md) that we have identified as the package hub has grown. Although meeting these checklist items is not a prerequisite to being added to the Hub, we have found that packages which don't conform provide a worse user experience. 
+
+### First run experience
+- [ ] The package includes a README which explains how to get started with the package and customise its behaviour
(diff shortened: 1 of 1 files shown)


dbeatty10 (approved):
Love adding this as a pull request template!

I'll be curious if the number of checklist items is manageable or overwhelming.

According to Amazon, 7,454 Kindle readers highlighted this section of the [Checklist Manifesto](https://www.amazon.com/Checklist-Manifesto-How-Things-Right/dp/0312430000/ref=sr_1_1?crid=1Y323L6DBLW4D&keywords=checklist+manifesto&qid=1670430464&s=books&sprefix=checklist+manifesto%2Cstripbooks%2C102&sr=1-1
):
> The checklist cannot be lengthy. A rule of thumb some use is to keep it to between five and nine items, which is the limit of working memory.

But since this is a **one-time** checklist (rather than a repeated one), it's quite possible that we'd rather have it be trending towards "comprehensive" rather than "minimal".

In the future, we may also want different PR templates for internal script changes vs. listing of packages.

Let's ship it as-is and then iterate as-needed. :shipit:

joellabes:
> I'll be curious if the number of checklist items is manageable or overwhelming.

I removed the MAY elements but yeah it's still beefy. Hoping that as a one-off it's not too arduous 

> we may also want different PR templates for internal script changes vs. listing of packages.

Excellent point!

PR #213, opened 2022-12-14
Add Snowplow's e-commerce package to the hub

emielver:
## Description 

Snowplow has created an e-commerce package for Snowplow users, which works in much the same way as the `web` and `mobile` packages we've had around for a while.

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -258,7 +258,8 @@
         "dbt-snowplow-web",
         "dbt-snowplow-mobile",
         "dbt-snowplow-media-player",
-        "dbt-snowplow-normalize"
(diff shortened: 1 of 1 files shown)


dbeatty10 (approved):
Awesome @emielver 🤩 

It should show up at https://hub.getdbt.com/snowplow/ in about an hour.

dbeatty10:
@emielver your new package is live on [hub.getdbt.com](https://hub.getdbt.com/):
image

PR #215, opened 2022-12-20
update hub.json

thutuva:
Added 2 packages and updated the repository name for 1:
2 packages added:  tuva_chronic_conditions and pmpm
1 package updated:  chronic_conditions to cms_chronic_conditions (still the same repository in Github as previously but renamed)

## Description 

Tell us about your new package! 
We are adding 2 new packages:
1.  tuva_chronic_conditions creates chronic conditions flags based on cohorts and logic we created internally at Tuva
2. pmpm calculates and provides the per member/per month cost of healthcare for a patient based on their claims data

I have also updated the name of 1 existing package:
1.  chronic_conditions is now cms_chronic_conditions.  The repository is the same as before but has been renamed.  Since we are releasing our own version of the chronic conditions logic we need to different/specify the package that contains the CMS logic.

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -273,11 +273,13 @@
     ],
     "tuva-health": [
         "the_tuva_project",
-        "chronic_conditions",
+        "cms_chronic_conditions",
         "claims_preprocessing",
         "data_profiling",
         "readmissions",
-        "terminology"
+        "terminology",
+        "pmpm",
+        "tuva_chronic_conditions"
     ],
     "yu-iskw": [
         "dbt-airflow-macros",


dbeatty10 (approved):
I'm approving this shortly, which will add the new packages to the Hub.

The "old" `chronic_conditions` will still show up until we do some things on our end to "archive" it.

Once I do the steps on our end, the Hub will re-route anybody that still has `chronic_conditions` listed in their packages.yml to the renamed `cms_chronic_conditions` package instead.

thutuva:
Thanks, Doug!  Happy Holidays!

dbeatty10:
Happy Holidays @thutuva !

image

I'll let you know once the "archive" step is completed for `chronic_conditions`.

dbeatty10:
✅ "Archive" step is done. This completes the rename from `chronic_conditions` to `cms_chronic_conditions `.

Before:
image

After:
image

PR #217, opened 2022-12-22
Update hub.json

fivetran-joemarkiewicz:
## Description 

The new Amazon Ads (and Source) packages developed by @fivetran-catfritz will follow the same nature of the existing Fivetran Ad packages and will roll up into the [Ad Reporting](https://github.com/fivetran/dbt_ad_reporting) dbt package. This package will produced modeled tables that leverage the Amazon Ads data from Fivetran's connector. It will enable customers to better understand the performance of their ads across varying grains: account, portfolio, campaign, ad group, keyword, and ad level reports.

Please be aware that the README and integration tests of the Amazon Ads package are still in development. They are likely to be completed the second week of January. However, I wanted to get ahead of my tasks and ensure I opened this PR sooner than later. Let me know if you have any questions! 😄 

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -202,7 +202,9 @@
         "dbt_amplitude",
         "dbt_amplitude_source",
         "dbt_recurly",
-        "dbt_recurly_source"
+        "dbt_recurly_source",
+        "dbt_amazon_ads",
+        "dbt_amazon_ads_source"
     ],
     "get-select": [
         "dbt-snowflake-monitoring"


dbeatty10 (approved):
> I wanted to get ahead of my tasks

@fivetran-joemarkiewicz -- always one step ahead 🏃 🏅

I'm approving this now. I don't _think_ it will publish to the Hub until you push your first version tags though. I guess we'll find out one way or the other!

dbeatty10:
Double-checked: can confirm that the hubcap script is successfully scanning for git tags once per hour, and it won't create the new package [here](https://hub.getdbt.com/fivetran/) until it finds some. 👍

PR #220, opened 2023-01-05
Add fhir-dbt-analytics package

jakubadamek:
## Description 

Data quality analytics for FHIR exported to BigQuery.

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -215,6 +215,9 @@
     "gitlabhq": [
         "snowflake_spend"
     ],
+    "google": [
+        "fhir-dbt-analytics"
+    ],
     "hightouchio": [
         "dbt-hightouch"
     ],


joellabes (approved):
@jakubadamek Thanks for submitting this! A couple of non-blocking comments: 

- Because you're pinned to specific patch versions of [dbt_expectations and dbt_unittest](https://github.com/google/fhir-dbt-analytics/blob/856445a15a9b010be2dae97c7a93d7d965790b5a/packages.yml#L16-L19), you may find that your users run into conflicts if they specify a different patch inside the same minor version. 
- Your readme specifies dbt Core v1.2 or above, but that's not defined in `require-dbt-version` in `dbt_project.yml`. 
- It's not out of the question that some users might already have a model called [metric](https://github.com/google/fhir-dbt-analytics/blob/856445a15a9b010be2dae97c7a93d7d965790b5a/models/post_processing/metric.sql) (for example), so you might find that users have a clash here. 

And a clarifying question for me - the readme specifies that it needs to be on dbt Core, is it incompatible with Cloud in some way or have you just not provided instructions for dbt Cloud? 

Welcome to the Hub!

jakubadamek:
Thanks a lot, Joe! I already fixed require-dbt-version and will fix the patch versions straight away. 
We will see if we want to rename `metric` and perhaps other models.
As to dbt Cloud - it's just the latter.

jakubadamek:
Would you like to merge this PR now, Joe? Thanks!

joellabes:
Yes that's perfect, thanks so much @jakubadamek!

PR #221, opened 2023-01-11
Added Datomni/profitwell_metrics and Datomni/ga4_metrics packages

enikovarga:
## Description 

[profitwell_metrics](https://github.com/Datomni/profitwell_metrics): this package provides daily and monthly metrics based on data generated by Profitwell. The package is compatible with BigQuery.

[ga4_metrics](https://github.com/Datomni/ga4_metrics): this dbt package sessionizes the ga4 dataset and creates a set of traffic metrics based on the sessions.

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -289,5 +289,8 @@
     "yu-iskw": [
         "dbt-airflow-macros",
         "dbt-unittest"
+    ],
+    "Datomni": [
+        "profitwell_metrics"
     ]
 }
--- a/hub.json
+++ b/hub.json
@@ -300,7 +300,8 @@
         "data_profiler"
     ],
     "Datomni": [
-        "profitwell_metrics"
+        "profitwell_metrics",
+        "ga4_metrics"
     ],
     "Aaron-Zhou": [
         "dbt-synapse-statistics"


dbeatty10 (changes requested):
Cool @enikovarga !

Could you explicitly specify `require-dbt-version` within `dbt_project.yml`? When this is specified, then dbt can proactively verify version compatibility when users install your package.

The relevant docs are [here](https://docs.getdbt.com/reference/project-configs/require-dbt-version), and [here](https://github.com/jpmmcneill/dbt-graph-theory/blob/c22353b1ff9e5c454e632ce15687d0521f3880b5/dbt_project.yml#L4) is a real-world example.

You can also specify only a lower bound without an upper bound if you want:
`require-dbt-version: ">=0.13.0"`

enikovarga:
@dbeatty10 thanks for the review! The `require-dbt-version` is now specified in the `dbt_project.yml` file.

dbeatty10:
@joellabes would you be available to take a look at this PR? If not, you can send it back to me.

See description for links to each package repo in GitHub.

joellabes (approved):
thank you @enikovarga and @dbeatty10!

PR #222, opened 2023-01-11
feat: add dbt-graph-theory package

jpmmcneill:
## Description 

Package to deal help when data you are dealing with can be modelled as a graph.

https://github.com/jpmmcneill/dbt-graph-theory

Examples of such things:
- events
- history tables

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -224,6 +224,9 @@
     "jitsucom": [
         "dbt-jitsu"
     ],
+    "jpmmcneill": [
+        "dbt-graph-theory"
+    ],
     "kristeligt-dagblad": [
         "dbt_ml"
     ],


dbeatty10 (approved):
Nice @jpmmcneill ! 🤩

Been looking forward to this since you first mentioned it was in the works.

Note: After approving and merging, I need to take a manual action for it to be live on [hub.getdbt.com](https://hub.getdbt.com/). If it's not live in about an hour, could you ping me? That would mean I either forgot to take the manual action or ran into some other issue during re-deployment.

jpmmcneill:
> Nice @jpmmcneill ! 🤩
> 
> Been looking forward to this since you first mentioned it was in the works.
> 
> Note: After approving and merging, I need to take a manual action for it to be live on [hub.getdbt.com](https://hub.getdbt.com/). If it's not live in about an hour, could you ping me? That would mean I either forgot to take the manual action or ran into some other issue during re-deployment.

Will do @dbeatty10, thank you!

jpmmcneill:
Hey @dbeatty10 fyi I'm not able to see it on the hub atm

jpmmcneill:
Ah I see this caused an exception (sorry 😇). Possibly these are related

dbeatty10:
No worries @jpmmcneill -- you didn't do anything wrong.

### Explanation

Although [hub.getdbt.com](https://hub.getdbt.com/) can display and serve prerelease tags like [1.0.0-rc1](https://github.com/dbt-labs/dbt-utils/releases/tag/1.0.0-rc1) (see [here](https://hub.getdbt.com/dbt-labs/dbt_utils/1.0.0-rc1/)), it can't currently handle the situation where there are _only_ prereleases and aren't any final versions yet.

So since your only tag right now is [0.1.0-b1](https://github.com/jpmmcneill/dbt-graph-theory/releases/tag/0.1.0-b1), it wouldn't deploy it to hub.getdbt.com.

### Next steps
If and when you publish a final tag like `0.1.0`, then we can get it working for hub.getdbt.com. As of right now, we won't be able to add your package until then.

Separately, I've got a draft PR https://github.com/dbt-labs/hubcap/pull/229 that would allow this to work in the future.

jpmmcneill:
Thanks @dbeatty10 , there were only a few small things (possibly allready solved - I can't remember) that I wanted to do before releasing a non beta version.

I'll follow up here when 0.1.0 is released.

On a separate note, is there a standard dbt like when it comes to version tags having the "v" prefix? I've seen packages doing both (eg. Fivetran have "v"s)

jpmmcneill:
Hey @dbeatty10 - FYI there are now two releases, `0.1.0` and `0.1.0-b1`.

dbeatty10:
> Hey @dbeatty10 - FYI there are now two releases, `0.1.0` and `0.1.0-b1`.

Awesome! We'll see if it works in about an hour.

> On a separate note, is there a standard dbt like when it comes to version tags having the "v" prefix? I've seen packages doing both (eg. Fivetran have "v"s)

Both with and without "v"s are allowed in the tag names. If there is a "v" prefix in the git tag, we'll just strip it off before determining if the remainder is valid according to the [SemVer 2.0 spec](https://semver.org/spec/v2.0.0.html#is-v123-a-semantic-version).

dbeatty10:
@jpmmcneill Your package is live 🎉 

https://hub.getdbt.com/jpmmcneill/dbt_graph_theory/latest/

jpmmcneill:
Wooo! Thanks @dbeatty10 😄

PR #226, opened 2023-01-11
Added dbt-segment-activity-schema to dbt hub

cdussud:
## Description 

This package creates two activity schema activities from a Segment source table: **viewed page**, and **started session**

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -238,7 +238,8 @@
         "dbt_product_analytics"
     ],
     "narratorai": [
-        "dbt-activity-schema"
+        "dbt-activity-schema",
+        "dbt-segment-activity-schema"
     ],
     "omnata-labs": [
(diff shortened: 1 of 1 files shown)


joellabes:
Hi @cdussud, there are some boxes that you ticked above but I can't find them in the code: 
- details of supported warehouses
- `require-dbt-version` details in `dbt_project.yml`
- you are casting columns as varchar instead of using the `dbt.type_string()` macros, so won't work for warehouses that don't use varchar terminology

These aren't necessary for release, so I am going to merge this, but encourage you to review and make any changes to ensure a good first run experience for people installing your package.

joellabes (approved):

PR #234, opened 2023-01-17
Add package dbt-synapse-statistics

Aaron-Zhou:
## Description 

Create statistic views for Azure Synapse dedicated sql pool.

## Checklist

### First run experience
- [ ✓] The package includes a README which explains how to get started with the package and customise its behaviour
- [ ✓] The README indicates which data warehouses/platforms are expected to work with this package

### Customisability
- [ ✓] The package uses ref or source, instead of hard-coding table references.
- [ ✓] do not assume database/schema names in sources.

#### Dependencies on dbt Core
- [ ✓] The package has set a supported `require-dbt-version` range in `dbt_project.yml`. Example: A package which depends on functionality added in dbt Core 1.2 should set its `require-dbt-version` property to `[">=1.2.0", "<2.0.0"]`.
There's no dependency on other package from dbt package hub.
### Interoperability
- [ ✓] The package does not override dbt Core behaviour in such a way as to impact other dbt resources (models, tests, etc) not provided by the package.
- [ ✓] The package uses the cross-database macros built into dbt Core where available, such as `{{ dbt.except() }}` and `{{ dbt.type_string() }}`.
- [ ✓] The package disambiguates its resource names to avoid clashes with nodes that are likely to already exist in a project. For example, packages should not provide a model simply called `users`.

### Versioning
- [ ✓] (Required): The package's git tags validates against the regex defined in [version.py](/hubcap/version.py)
- [ ✓] The package's version follows the guidance of Semantic Versioning 2.0.0. (Note in particular the recommendation for production-ready packages to be version 1.0.0 or above)

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -298,5 +298,8 @@
     ],
     "jmangroup": [
         "data_profiler"
-    ]
+    ],
+    "Aaron-Zhou": [
+        "dbt-synapse-statistics"
+        ]
 }


dbeatty10:
https://github.com/Aaron-Zhou/dbt-synapse-statistics

Looks like a handy package for Synapse @Aaron-Zhou !

Could you go through the checklist in this PR and tick off all the boxes that are applicable to your package?

Aaron-Zhou:
> https://github.com/Aaron-Zhou/dbt-synapse-statistics
> 
> Looks like a handy package for Synapse @Aaron-Zhou !
> 
> Could you go through the checklist in this PR and tick off all the boxes that are applicable to your package?

I updated the PR comments

dbeatty10 (approved):
Awesome, @Aaron-Zhou!

If all goes well, it should be up on the dbt Package Hub within the next few of hours.

If not, please ping me so we can troubleshoot.

dbeatty10:
@Aaron-Zhou you are not only live on the [dbt Package Hub](https://hub.getdbt.com/), you now occupy the top spot in our (alphabetized) list!
image

PR #239, opened 2023-02-07
add Education Analytics packages

ejoranlienea:
## Description 

The EDU framework creates a data warehouse model on top of the open source [Ed-Fi](https://www.ed-fi.org/) data standard for K-12 education.

https://github.com/edanalytics/edu_edfi_source
https://github.com/edanalytics/edu_wh

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -308,5 +308,9 @@
     ],
     "Aaron-Zhou": [
         "dbt-synapse-statistics"
+    ],
+    "edanalytics": [
+        "edu_edfi_source",
+        "edu_wh"
     ]
 }


dbeatty10 (commented):
Thanks for opening @ejoranlienea -- looks like some great stuff !

I noticed that your readme says:
> dbt version required: >=1.0.0, <2.0.0

Is there any reason why you wouldn't want to add the following to both of your `dbt_project.yml` files (as shown [here](https://docs.getdbt.com/reference/project-configs/require-dbt-version#pin-to-a-range))?
```
require-dbt-version: [">=1.0.0", "<2.0.0"]
```

ejoranlienea:
@dbeatty10 
Thank you, this has been added to main in both repos and will be included in our next release.
https://github.com/edanalytics/edu_edfi_source/blob/e5c70996a6a792f61f723046bb933876ef13c4cc/dbt_project.yml#L5
https://github.com/edanalytics/edu_wh/blob/2e74c5e12c617c3308de7103be85322a0696343e/dbt_project.yml#L5

dbeatty10 (approved):

dbeatty10:
@ejoranlienea your new packages are live at https://hub.getdbt.com/:

image

PR #240, opened 2023-02-08
Add dbt-snowflake-query-tags package

NiallRees:
## Description 

Link to your package's repository: https://github.com/get-select/dbt-snowflake-query-tags

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
- [ ] The package has set a supported `require-dbt-version` range in `dbt_project.yml`. Example: A package which depends on functionality added in dbt Core 1.2 should set its `require-dbt-version` property to `[">=1.2.0", "<2.0.0"]`. - Not needed
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -210,7 +210,8 @@
         "dbt_amazon_ads_source"
     ],
     "get-select": [
-        "dbt-snowflake-monitoring"
+        "dbt-snowflake-monitoring",
+        "dbt-snowflake-query-tags"
     ],
     "gitlabhq": [
         "snowflake_spend"


dbeatty10 (approved):

joellabes:
This is a very cool package! He's done it again folks 🚀

dbeatty10:
@NiallRees your new package is live at https://hub.getdbt.com/:

image

NiallRees:
> Cool @NiallRees !
> 
> Is there a minimum version of dbt that you'd expect to work? Or would you expect it to work with "all" versions of dbt? I saw you said "Not needed", but just want to double-check.

I don't know of any it wouldn't work with, which is why I went for not needed. It seems like the query tag interface in dbt-snowflake has been pretty consistent for a while.

NiallRees:
> @NiallRees your new package is live at https://hub.getdbt.com/:
> 
> image

Thank you!!

PR #243, opened 2023-02-12
add dbt-duckdb-utils

sdebruyn:
## Description 

Utility package for users of duckdb to map macros from popular dbt packages. I'm using duckdb in a personal project and I'm fixing the ones I bump into.

Link to your package's repository: https://github.com/sdebruyn/dbt-duckdb-utils

### First run experience

### Customisability

#### Dependencies on dbt Core

### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -265,6 +265,9 @@
         "dbt-customer-journey-analysis",
         "dbt-events-registry"
     ],
+    "sdebruyn": [
+        "dbt-duckdb-utils"
+    ],
     "sgoley": [
         "dbt-postgres-utils"
     ],


joellabes:
Thanks @sdebruyn! I see you've got hooks into dbt_utils as well - we're _hoping_ that by moving a lot of the cross-db stuff into the adapters, you should never* have to provide an override for any utils macro. If you come across any, let us know in the utils repo - even if we can't fix it directly, it's useful to know what sorts of things cause issues for other adapters. 

*well, never-ish

joellabes (approved):

PR #252, opened 2023-04-04
Add dbt_tld for maintenance of top level domain lookup table

GJMcClintock:
## Description 

This is a simple package that maintains a seed file of the currently accepted IANA TLDs.

Link to your package's repository: https://github.com/GJMcClintock/dbt_tld/tree/0.1.0

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -240,6 +240,9 @@
     "gitlabhq": [
         "snowflake_spend"
     ],
+    "GJMcClintock": [
+        "dbt_tld"
+    ],
     "google": [
         "fhir-dbt-analytics"
     ],


joellabes:
Hey @GJMcClintock, this is very cool! One comment: instead of calling `split_part` directly, you should use the [cross-db implementation defined here](https://docs.getdbt.com/reference/dbt-jinja-functions/cross-database-macros#split_part), i.e. `{{ dbt.split_part() }}`. This will ensure that any warehouses that use a different syntax are also compatible. 

That can be handled in a different release though, so I'll merge this!

joellabes (approved):

GJMcClintock:
@joellabes - I just released a patch to switch to the cord split_part - I didn't realize that was a function. Thanks for the heads up!

PR #254, opened 2023-04-27
edit version numnber

tuvaforrest:
Hi, I made a mistake with version numbers on our package, and while we've fixed the issue in github, it still exists in dbthub.

for the [tuva-health/the_tuva_project](https://github.com/tuva-health/the_tuva_project) We accidently release v1.3.0 instead of v0.3.0.  We deleted the release in github, but it still shows up in dbt hub.

Apologies.  Can you help remove 1.3.0 from dbt hub so it reflects github?

Thanks!

## Description 

## Checklist

### First run experience
- [ ] The package includes a README which explains how to get started with the package and customise its behaviour
- [ ] The README indicates which data warehouses/platforms are expected to work with this package

### Customisability
- [ ] The package uses ref or source, instead of hard-coding table references.
- [ ] do not assume database/schema names in sources.

#### Dependencies on dbt Core
### Interoperability
- [ ] The package does not override dbt Core behaviour in such a way as to impact other dbt resources (models, tests, etc) not provided by the package.

### Versioning
- [ ] (Required): The package's git tags validates against the regex defined in [version.py](/hubcap/version.py)

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -338,7 +338,7 @@
         "dbt-activity-schema"
     ],
     "tuva-health": [
-        "the_tuva_project",
+        "the_tuva_project", 
         "cms_chronic_conditions",
         "claims_preprocessing",
         "data_profiling",


dbeatty10:
I see what you are talking about here:
image

I'm going to revert https://github.com/dbt-labs/hub.getdbt.com/pull/2443 in order to remove / "yank" the accidental 1.3.0 version.

I haven't tried the "Revert" button in GitHub before, so I'll see if that works:
image

dbeatty10:
Actually, won't be able to do a revert in this case because there is a non-Tuva package in the same PR:
https://github.com/dbt-labs/hub.getdbt.com/pull/2443/files

Will need to create a new PR to update the relevant files instead.

dbeatty10:
It's [all good again](https://hub.getdbt.com/tuva-health/the_tuva_project/latest/) @tuvaforrest:

image

I'm going to close this as resolved, but please reach out if you need any other assistance with this.

tuvaforrest:
Hi, 

It looks like this change has reverted.  I'm not sure if hubcap is picking up the deleted release somehow, or if its just a vestige from when the release existed?  Any ideas on how to fix?

And apologies, not sure the right process here, I can put in a pr on [hub.getdbt.com](https://github.com/dbt-labs/hub.getdbt.com) if you would like, but again not sure if the change would stick.

Thanks again

PR #255, opened 2023-05-01
Issue with deleted release

tuvaforrest:
Hi, I accidently released version 1.3.0 instead of 0.3.0 and that version was picked up by hubcap.  I opened a pr and it was resolved [here](https://github.com/dbt-labs/hubcap/pull/254) by @dbeatty10 , but it seems like that change has been "reverted," or hubcap has reupdated the version to 1.3.0 even though that release has been deleted?

Let me know how to help fix. Thanks again,
Forrest

## Description 

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -338,7 +338,7 @@
         "dbt-activity-schema"
     ],
     "tuva-health": [
-        "the_tuva_project",
+        "the_tuva_project", 
         "cms_chronic_conditions",
         "claims_preprocessing",
         "data_profiling",


tuvaforrest:
I may have found the issue, or part of it: when we deleted the release, we didn't also delete the tag.  I've just deleted the tag, so hopefully the same change will fix it permanently this time.

dbeatty10:
Yep, hubcap will eagerly pick up any tags in GitHub when it runs at the top of every hour.

I reverted the most recent PR that added `1.3.0` within https://github.com/dbt-labs/hub.getdbt.com/pull/2449, and it looks like [all is well](https://hub.getdbt.com/tuva-health/the_tuva_project/latest/) again:

image

I'm going to close this, but let us know if this pops up again and we'll take another look.

PR #257, opened 2023-05-12
Add a Snowflake Package dbt-incremental-stream

arnoN7:
add Snowflake Incremental Stream package to hub

## Description 
**This dbt package is for Snowflake ❄️ only.**

**It is reproducing dbt incremental materialization** leveraging on Snowflake [streams](https://docs.snowflake.com/en/user-guide/streams-intro) to :
* **Improve model performance 💨**. [Stream](https://docs.snowflake.com/en/user-guide/streams-intro) accelerate source table scan focusing new data only 
* **Optimise FinOps 💰**. Snowflake warehouse will be started only if [stream](https://docs.snowflake.com/en/user-guide/streams-intro) is not empty. 
* Simplify code. `is_incremental()` is generally not required as [stream](https://docs.snowflake.com/en/user-guide/streams-intro) natively collect "new" rows

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -2,6 +2,9 @@
     "Aaron-Zhou": [
         "dbt-synapse-statistics"
     ],
+    "arnoN7": [
+        "dbt-incremental-stream"
+    ],
     "Datavault-UK": [
         "dbtvault"
     ],


dbeatty10 (approved):
Cool @arnoN7 ! ❄️

dbeatty10:
https://hub.getdbt.com/arnon7/incr_stream/latest/

image

PR #268, opened 2023-05-20
Update hub.json to use sutrolabs organization

Stephen986:
I previously put our package under 'census', but the github org is 'sutrolabs'.  The main PR was already approved.

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -70,9 +70,6 @@
         "dbt-date",
         "dbt-expectations"
     ],
-    "census": [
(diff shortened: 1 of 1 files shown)


dbeatty10 (approved):

dbeatty10:
@Stephen986 After merging this, I re-deployed the hubcap script to Heroku with your changes incorporated. So this should be live at hub.getdbt.com in a little less than an hour. If not, let us know and we can tackle it again on Monday.

PR #271, opened 2023-06-09
Update hub.json

zshandy:
## Description 

This is a tool for creating the column level lineage graph for dbt utilizing a current dbt project. This tool can create an interactive graph on a webpage to explore the column level lineage for the inputted models, it works with the Postgres database connection, and other connection types or dialects are under development.

https://github.com/sfu-db/dbt-lineagex

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -355,5 +355,8 @@
     ],
     "datnguye": [
         "dbt-translate"
+    ],
+    "sfu-db": [
+        "dbt-lineagex"
     ]
 }


joellabes:
Hey @zshandy, this looks fascinating! 

It doesn't seem to have any components of a dbt project though, is that right? I.e. it doesn't create any models in your warehouse, or provide any macros etc. 

The package hub is for [packages that extend a dbt project](https://docs.getdbt.com/docs/build/packages) - it looks like this should be submitted to PyPI instead? 

If you wanted to share more information about what you've made, you could definitely share it on the community forum: https://discourse.getdbt.com/c/show-and-tell/22

lmk if I'm misinterpreting!

zshandy:
Hey @joellabes ,

Yes you are right, it doesn't create any models or macros. The core this uses(https://github.com/sfu-db/lineagex) is in fact on Pypi. However, I just made this as a dbt wrapper such that the user only needs to add one line in the packages.yml and then run a bash file in the folder for producing the output. I thought it would be more convenient for producing the output that way than pip install, turn up python and then call the function. 

Do you think it is still possible to be published on the hub? Or maybe the forum is the best place for advertising it?

Thank you very much!

PR #272, opened 2023-06-20
Update hub.json - add Aidbox/dbt_fhir package

Aitem:
## Description 

This dbt package provides useful macros, build in models, and tests for working with Aidbox FHIR platform. Only the PostgreSQL data source is supported.

Link to your package's repository: https://github.com/Aidbox/dbt_fhir

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -2,6 +2,9 @@
     "Aaron-Zhou": [
         "dbt-synapse-statistics"
     ],
+    "Aidbox": [
+        "dbt_fhir"
+    ],
     "arnoN7": [
         "dbt-incremental-stream"
     ],


Aitem:
@joellabes

Aitem:
Hi @joellabes @dbeatty10. Any updates on this ?

joellabes (approved):
@Aitem we can approve this, but some comments: 
- you aren't using sources, which means your table lineage won't be complete 
- it looks like the content of the [sample project](https://github.com/Aidbox/dbt-sample-project/blob/main/models/views/dim_location.sql) should probably be incorporated into the main package, so that the data is consistently modelled from the outset instead of needing to copy a separate sample project in.

PR #273, opened 2023-06-20
Update hub.json

aneiderhiser:
Added 3 new packages to tuva-health.  These were listed previously but then we took them down for a period while we made updates.  We're now ready to put them back up.

---

## Description 

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability
- [ ] The package does not override dbt Core behaviour in such a way as to impact other dbt resources (models, tests, etc) not provided by the package.

### Versioning
- [ ] (Required): The package's git tags validates against the regex defined in [version.py](/hubcap/version.py)

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -347,7 +347,10 @@
         "dbt-activity-schema"
     ],
     "tuva-health": [
-        "the_tuva_project"
+        "the_tuva_project",
+        "the_tuva_project_demo",
+        "medicare_cclf_connector",
+        "medicare_lds_connector"
     ],
     "yu-iskw": [
         "dbt-airflow-macros",


aneiderhiser:
@dbeatty10

dbeatty10:
@aneiderhiser I'd like to treat each of these as net-new packages so we can make sure each has gone through our checklist.

I've added the checklist to the PR description. Can you duplicate the checklist and fill it out for each new package you are adding? Then ping us again once that's done?

I'm _guessing_ we won't want to add https://github.com/tuva-health/the_tuva_project_demo as a package on the Hub since it looks like it is merely a demo and not a package with reusable macros or transformations. I didn't look at this in-depth, so please let me know if it's more than a demo.

aneiderhiser:
@dbeatty10 sounds great I'll do that.  2 of the things we want to add are actually projects, not packages.  Do you allow projects on dbt hub?

dbeatty10:
> 2 of the things we want to add are actually projects, not packages. Do you allow projects on dbt hub?

The dbt Package Hub is intended only for dbt packages that contain macros or models that are intended to be shared across multiple dbt projects (via `packages.yml` and `dbt deps`).

Regular dbt projects can just be shared on GitHub. Then people can use those projects by cloning and building them. Demo projects is a perfect example of this.

aneiderhiser:
@dbeatty10 Makes sense.  Then I think the only package we want on dbt hub is the_tuva_project.  We can remove medicare_cclf_connector, which is a dbt project, not a package.  I'll submit a PR for this.

PR #274, opened 2023-06-28
Update hub.json

fivetran-joemarkiewicz:
## Description 

Adding our soon to be released package from this quarter!
- [Sap](https://github.com/fivetran/dbt_sap) and [Sap Source](https://github.com/fivetran/dbt_sap_source)
- [Twilio](https://github.com/fivetran/dbt_twilio) and [Twilio Source](https://github.com/fivetran/dbt_twilio_source)
- [Qualtrics](https://github.com/fivetran/dbt_qualtrics) and [Qualtrics Source](https://github.com/fivetran/dbt_qualtrics_source)

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -237,7 +237,13 @@
         "dbt_reddit_ads",
         "dbt_reddit_ads_source",
         "dbt_recharge",
-        "dbt_recharge_source"
+        "dbt_recharge_source",
+        "dbt_sap_source",
+        "dbt_sap",
+        "dbt_twilio_source",
+        "dbt_twilio",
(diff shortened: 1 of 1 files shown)


joellabes (approved):
Looking good! I did notice that your badge says v1.0 works but your dbt_project.yml requests 1.3 and above
image

fivetran-joemarkiewicz:
@joellabes thanks for calling that out! Will update the badge before the initial release

PR #276, opened 2023-07-04
Check both `packages.yml` + `dependencies.yml` for `packages` deps

### LINKED ISSUE #275: Hubcap needs to know about dependencies.yml
https://github.com/dbt-labs/dbt-core/issues/7631 / https://github.com/dbt-labs/dbt-core/issues/7372 added the ability for package dependencies to be defined in `dependencies.yml`. 

Hubcap needs to know to check that file's `packages` block as well. 

It also needs to enforce the [rules/outcomes listed here](https://github.com/dbt-labs/dbt-core/issues/7631#issuecomment-1564965453), especially the rule about only one of `packages.yml` and `dependencies.yml` being allowed to contain a `packages` key. I suspect that will need to be through duplication of logic as opposed to being able to just use dbt-core's logic, which is unfortunate.

jtcohen6:
resolves #275

### DIFF
--- a/hubcap/package.py
+++ b/hubcap/package.py
@@ -41,6 +41,12 @@ def parse_pkgs(repo_dir):
         with open(repo_dir / Path("packages.yml"), "r") as stream:
             pkgs = yaml.safe_load(stream)
             return pkgs.get("packages", []) if pkgs else []
+    # new in v1.6
+    # a project can declare 'packages' in *either* packages.yml or dependencies.yml, not both
+    elif os.path.exists(repo_dir / "dependencies.yml"):
+        with open(repo_dir / Path("dependencies.yml"), "r") as stream:
+            pkgs = yaml.safe_load(stream)
+            return pkgs.get("packages", []) if pkgs else []
     else:
         return []
 


hubcap/package.py
```diff
@@ -41,6 +41,12 @@ def parse_pkgs(repo_dir):
         with open(repo_dir / Path("packages.yml"), "r") as stream:
             pkgs = yaml.safe_load(stream)
             return pkgs.get("packages", []) if pkgs else []
+    # new in v1.6
+    # a project can declare 'packages' in *either* packages.yml or dependencies.yml, not both
+    elif os.path.exists(repo_dir / "dependencies.yml"):
```
joellabes, line 46: I was going to note that this would silently fail if they did put something in both, but then I remembered that if they did that their project would fail to compile in the first place. So assuming people run their packages before submitting them to the hub, we can be a lil sloppier here and be confident that it'll be ok.

joellabes (approved):
🏎️💨

MichelleArk (approved):

PR #279, opened 2023-08-09
Add `yu-iskw/dbt-gcp-biling`

yu-iskw:
## Description 
I implemented a dbt Package to deal with Google Cloud billing exported to BigQuery.

Link to your package's repository: https://github.com/yu-iskw/dbt-gcp-billing

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -365,7 +365,8 @@
     ],
     "yu-iskw": [
         "dbt-airflow-macros",
-        "dbt-unittest"
+        "dbt-unittest",
+        "dbt-gcp-billing"
     ],
     "datnguye": [
         "dbt-translate"


yu-iskw:
@joellabes @dbeatty10 Can you please take a look?

joellabes (approved):
This is cool! You might want to consider adding models that query your sources (even if it's just a `select *`) so that end users don't need to copy-paste `{{ source(var('dbt-gcp-billing::detailed_usage_cost::dataset'), "dbt_gcp_billing__detailed_usage_cost") }}` around, but that's not blocking!

yu-iskw:
@joellabes That's a good idea. I will add staging models to simplify them for end users. Thank you!

PR #280, opened 2023-08-31
feat: source_db package

BfdCampos:
## Description 

Link to your package's repository: https://github.com/BfdCampos/source_db/

## Checklist

### First run experience
- [ ] ~The README indicates which data warehouses/platforms are expected to work with this package~

### Customisability
- [ ] ~The package uses ref or source, instead of hard-coding table references.~

#### Dependencies on dbt Core
### Interoperability
- [ ] ~The package uses the cross-database macros built into dbt Core where available, such as `{{ dbt.except() }}` and `{{ dbt.type_string() }}`.~

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -8,6 +8,9 @@
     "arnoN7": [
         "dbt-incremental-stream"
     ],
+    "BfdCampos": [
+        "source_db"
+    ],
     "Datavault-UK": [
         "dbtvault"
     ],


BfdCampos:
Hi @joellabes, thanks so much for reviewing and reaching out! 

So this package is designed to dynamically set the database source based on an environment variable. They are useful for conditional database routing but work differently from the defer feature from what I understand from the [documentation](https://docs.getdbt.com/reference/node-selection/defer).

The `source_db` package allows the user to easily switch between databases on the fly based on environmental variables which can be declared with your `dbt` command or at a global environment level for a multi-tenant db setup. We find a lot of use for this in my own company for when we need to test a run locally based off of data in other databases (`dev` or `prod`). This is particularly useful for us as we cannot run `dbt` in production at all (meaning no access to the production artefacts). Only via automated systems.

From what I understood, the `defer` feature is for particularly useful for optimising computational resources in CI. It switches between databases or schemas based on the existence of a model in the current environment, automatically referring to a production model if a development one does not exist, but requires that a manifest from a previous `dbt` invocation be passed to the `--state` flag or env var which for us, is not possible. So this feature does not work for our use case unfortunately.

In summary, the `source_db` package allows for conditional logic based on environment variables, enabling more complex routing that `defer` does not offer at this time.

I would also be happy if this feature was taken up by `dbt` as a default command if you think it adds value 😊  

(Anecdotal but the main reason why I even made this PR was because I've already sent the exact copy of the code for the package to 4 friends at different companies because they found it useful, so I thought instead of having to send it manually, why not make it into a package?)

joellabes:
Works for me! It might be worth adding a bit more of that context to your readme to help folks understand the contexts where it's useful, but let's get this merged!

joellabes (approved):

BfdCampos:
Perfect will do now!! Thanks so much @joellabes 🙌🙌

BfdCampos:
@joellabes apologies for the delay. This week has taken the better of me. But I have added the notes to the `README` as you suggested and made another release so that the package has the latest information. Thanks 😊

PR #289, opened 2023-10-02
re-add bfdcampos/source_db

joellabes:
## Description 

Link to your package's repository: https://github.com/bfdcampos/source_db

Second attempt at adding package after needing to revert #280 because it seemed to crash hubcap for some reason


joellabes:
@dbeatty10 Every time I merge this, Hubcap stops running and detecting other package updates, so I have reverted the PR. Any ideas why this might be? It seems to have the expected components in dbt_project.yml etc. 

cc @BfdCampos (sorry!)

BfdCampos:
No worries @joellabes! Having a look as well in case there was something I did wrong but really I can't seem to find anything that's sticking out.

I did find [an example package that was added](https://github.com/techindicium/dbt-metalog/blob/main/dbt_project.yml) a few months ago that only described the `macro-paths`. Whereas mine described all the rest `test-path`, `log-path`, etc.

That's the only major thing I can think of but please do let me know if there's anything I can do from my side to help 😊

BfdCampos:
Hi @joellabes @dbeatty10 

Been a while but just wanted to check if we have capacity to see if we can merge this package? We're currently building multiple dbt projects for our teams and would be keen to import this as a package rather than a custom individual macro for each project.

Thanks.

dbeatty10:
@joellabes I don't know why it is crashing hubcap 😨 

I'd suggest running hubcap in a test mode as described [here](https://github.com/dbt-labs/hubcap/blob/main/CONTRIBUTING.md#run-in-test-mode) and see if it crashes or not.

joellabes:
@BfdCampos I worked it out! The [0.1.0 release](https://github.com/BfdCampos/source_db/tree/v0.1.0) of the package is incorrectly structured, and hubcap can't find a dbt_project.yml file. I think if you delete that release + tag from GitHub then it should be OK

joellabes:
Hoping that this will work now!

PR #296, opened 2024-01-03
Update hub.json

fivetran-joemarkiewicz:
## Description 

These new dbt packages model the respective data synced by the Fivetran connectors into analysis ready tables.

 - https://github.com/fivetran/dbt_servicenow
 - https://github.com/fivetran/dbt_workday
 - https://github.com/fivetran/dbt_salesforce_marketing_cloud

Please be aware these packages have not yet been released. The initial release is slated for our upcoming sprint. They should all have the v0.1.0 release before the end of January.

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -275,7 +275,10 @@
         "dbt_twilio_source",
         "dbt_twilio",
         "dbt_qualtrics_source",
-        "dbt_qualtrics"
+        "dbt_qualtrics",
+        "dbt_workday",
+        "dbt_servicenow",
+        "dbt_salesforce_marketing_cloud"
     ],
     "fullstorydev": [
         "dbt_fullstory"


dbeatty10 (approved):
🚀 3 new packages to celebrate January 3.

dbeatty10:
@fivetran-joemarkiewicz both should show up on the Hub within the next 30 minutes. Could you let me know if they don't show for some reason?

fivetran-joemarkiewicz:
Thanks so much @dbeatty10! We actually have not yet released these packages, but plan to at the end of January. I believe with this update the packages should be picked up once we cut the new releases in about two weeks time. Thanks again!

PR #298, opened 2024-01-14
adding gmdata-co to hub.json

jeff-skoldberg-gmds:
## Description 

This package is the first in a series of packages to come that aim at making connectors from Portable.io easier to use. 
This packages contains a single macro that will unnest the json payload from Portable's Google Sheets connector.
It transforms the single variant column called "ROWDATA" back to the original columns as they existed in the Google Sheet.

Since this is my first time doing this, I'm not 100% clear on how this all works.  I guessed the naming convention in hub.json is:
org-name: [repo-name]... but that was not clear to me from the docs.  I guess somehow it automatically will link to my repo based on that?

I have questions about versioning.  Can you hit me up on Slack so I can ask?  I'm Jeff Skoldberg on dbt Slack.
Can I update the readme without updating the version?  Example, I will need to update the install instructions in the readme if this is going live.

https://github.com/gmdata-co/dbt-portable-google-sheets

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -290,6 +290,9 @@
     "gitlabhq": [
         "snowflake_spend"
     ],
+    "gmdata-co": [
+        "dbt-portable-google-sheets"
+    ],
     "google": [
         "fhir-dbt-utils",
         "fhir-dbt-analytics"


joellabes (approved):

joellabes:
lgtm! I've sent you a slack message re your versioning questions

PR #301, opened 2024-01-16
Add `dbt-assertions` package to hub

AxelThevenot:
## Description 

> dbt-assertions is a package designed for data testing capabilities at the row level integrated into the data pipeline flagging each row based on specific assertions it fails.

Link to your package's repository:  https://github.com/AxelThevenot/dbt-assertions

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -5,6 +5,9 @@
     "Aidbox": [
         "dbt_fhir"
     ],
+    "AxelThevenot": [
+        "dbt-assertions"
+    ],
     "Datavault-UK": [
         "dbtvault"
     ],


AxelThevenot:
Update for a generic test and better documentation in `0.1.1`
Everything ok now 😇

@joellabes , @dbeatty10

joellabes (changes requested):
This is wildly cool, I love it 🤩 

Two comments: 
- now that this will be on the hub, please update the readme's sample code to a hub reference instead of installing directly from git
- the dbt-assertions image header is clever, but I think it runs afoul of our [trademark guidelines](https://www.getdbt.com/trademark-guidelines) as it's "combined with another logo". Can you also update that? 

ty!

AxelThevenot:
Oh my bad, let me change this during the day. I though it was a good idea but you're right

Thank you for the feedback I return to you as fast as possible

AxelThevenot:
What about this image if I keep the `for dbt` part ?
Good for you ? 
I'll update the documentation according to your answer, thank you 😄 

image

AxelThevenot:
What about this image if I keep the `for dbt` part ?
Good for you, @joellabes  ? 
I'll update the documentation according to your answer, thank you 😄 

image

joellabes:
Thanks for the quick turnaround @AxelThevenot! I have checked internally, and the takeaway is that you can't use the logo at all - the relevant part of the trademark guidelines is: 
>You may use the word Marks, but not the logo Marks, to truthfully describe the relationship between your software or services and ours

However you could do something like this: 
image

(as long as the word mark includes the ™, which your current version does)

or use the name `dbt-assertions` on its own because of this carve out: 
> dbt packages [...] may use the word mark “dbt” as part of the project name

image

Hopefully that makes sense - lmk if not, either here or `@joellabes` on the community Slack

AxelThevenot:
Everything is now updated (image and install) 
Thank you very much for your feedbacks !

joellabes (approved):

joellabes:
YAY I am so excited to take this for a spin!

PR #308, opened 2024-02-20
included metaops package for camunda

akozichev:
## Description 

DBT Package with the model for Camunda history

Link to your package's repository: https://github.com/metaops-solutions/dbt-camunda

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -335,6 +335,9 @@
     "lalalilo": [
         "athena-utils"
     ],
+    "metaops-solutions": [
+        "dbt-camunda"
+    ],
     "mjirv": [
         "dbt-datamocktool",
         "dbt_product_analytics"


joellabes (changes requested):
Hey @akozichev, thanks for submitting this! 

I noticed an issue while doing the review. To ensure your users don't have a bad experience, I suggest you provide a way to specify an alternative schema location than `public`, e.g. vars. 
One warning on this note - your README states that source tables are expected to be in `camunda_history`, but the [`schema` config here](https://github.com/metaops-solutions/dbt-camunda/blob/main/models/staging/schema.yml#L5) will [take priority](https://docs.getdbt.com/reference/resource-properties/schema)

Please also move your block in `hub.json` to be alphabetically sorted (this is the only mandatory thing before merge). 

Thanks!

akozichev:
Hi @joellabes, can you have a look, please?

akozichev:
can this be approved now? I have fixed the problems.

joellabes (approved):

PR #322, opened 2024-06-04
Use the timestamp of the first file within the GZIP archive for `published_at`

aranke:
## Description 

Use the timestamp of the first file within the GZIP archive for `published_at` instead of hard-coding to `1970-01-01`

Sample PR: https://github.com/dbt-labs/hub.getdbt.com-test/pull/251/files


joellabes:
This was intentionally removed in https://github.com/dbt-labs/hubcap/pull/141 because it led to a ridiculous number of redundant PRs being opened, what makes you want to bring it back? (Your approach is admittedly much more idempotent than just using the current time). 

Not _averse_ to it, but we don't display it anywhere so I'm not sure what filling it out solves.

aranke:
@joellabes I want to avoid errors like this one: https://github.com/dbt-labs/dbt-core/pull/10260. Currently, each version bump of `dbt-utils` will break `dbt-core` because the `deps` test will break.

Another option if we don't want to go this route is for the Core team to create new packages just for this test that we solely control.

aranke:
Furthermore, because we're using the idempotent release time, we should not run into merge conflicts like this one: https://github.com/dbt-labs/hub.getdbt.com/issues/1484#issuecomment-1142863530

joellabes:
We've discussed this separately and are closing for now

PR #324, opened 2024-07-26
Add dbt_eda_tools

shankararul:
# Description
A medley of macros that could be handy for your data exploration in DBT. The first iteration implements **Get Missing Dates**

✅ Get Missing Dates
`Finds all the missing dates in a model for the specified dimensions and filters according to the time granularity expected`

🚧 Fill Missing Dates (Coming soon)
`Fills the missing dates in a model for the specified dimensions and filters according to the time granularity expected`

🚧 Show as Percentage (Coming soon)
`Shows the value as percentage of the total value for the specified aggregations`

🚧 Exploratory data analysis (Coming soon)

> Numeric column exploration
`Get summary statistics such as Min, Max, Median, Null values, Percentiles, Standard deviation, etc. for numeric columns`

> Categoric column exploration
`Get summary statistics such as Count, Unique values, Null values for categoric columns`

> Timeseries column exploration
`Get summary statistics such as Start date, End date, granularity of the timeseries (day,month,year), null values, missing dates for timeseries columns`

Docs site:  https://shankararul.github.io/dbt_eda_tools/#!/macro/macro.dbt_eda_tools.get_missing_date
Link to your package's repository: https://github.com/shankararul/dbt_eda_tools

# Checklist
This checklist is a cut down version of the [best practices](https://github.com/dbt-labs/hubcap/blob/main/package-best-practices.md) that we have identified as the package hub has grown. Although meeting these checklist items is not a prerequisite to being added to the Hub, we have found that packages which don't conform provide a worse user experience.

## First run experience
 
## Customisability
### Packages for data transformation (delete if not relevant):
- [x]  provide a mechanism (such as variables) to customise the location of source tables.
- [x] do not assume database/schema names in sources..

### Dependencies on dbt Core
- [x]  The package has set a supported require-dbt-version range in dbt_project.yml. Example: A package which depends on functionality added in dbt Core 1.2 should set its require-dbt-version property to [">=1.2.0", "<2.0.0"].
### Dependencies on other packages defined in packages.yml:
 - [x]  Dependencies are imported from the dbt Package Hub when available, as opposed to a git installation.
 - [x]  Dependencies contain the widest possible range of supported versions, to minimise issues in dependency resolution.
 - [x]  In particular, dependencies are not pinned to a patch version unless there is a known incompatibility.
### Interoperability
  - [x]  The package does not override dbt Core behaviour in such a way as to impact other dbt resources (models, tests, etc) not provided by the package.
 - [x]  The package uses the cross-database macros built into dbt Core where available, such as {{ dbt.except() }} and {{ dbt.type_string() }}.
 - [x]  The package disambiguates its resource names to avoid clashes with nodes that are likely to already exist in a project. For example, packages should not provide a model simply called users.
## Versioning
  - [x]  (Required): The package's git tags validates against the regex defined in [version.py](https://github.com/hubcap/version.py)
  - [x]  The package's version follows the guidance of Semantic Versioning 2.0.0. (Note in particular the recommendation for production-ready packages to be version 1.0.0 or above)

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -87,7 +87,7 @@
         "datavault4dbt"
     ],
     "shankararul": [
-        "dbt_utils_medley"
+        "dbt_eda_tools"
     ],
     "Snowflake-Labs": [
         "dbt_constraints"
--- a/hub.json
+++ b/hub.json
@@ -86,6 +86,9 @@
     "ScalefreeCOM": [
         "datavault4dbt"
     ],
+    "shankararul": [
+        "dbt_utils_medley"
+    ],
     "Snowflake-Labs": [
         "dbt_constraints"
     ],


joellabes:
@shankararul this is so cool! My only concern is the overlap of calling it dbt_utils_medley when it's not related to dbt Labs' dbt_utils package. If you tweaked the name to something like dbt_medley or [dbt_eda_tools](https://en.wikipedia.org/wiki/Exploratory_data_analysis) or similar, I'd love to get this merged 🎉

joellabes (approved):

PR #330, opened 2024-10-16
Added dbt-dlt-Shopify package

dat-a-man:
## Description 

The Shopify dbt package offers data models to help you transform and analyze Shopify data .It's designed to integrate seamlessly with the dlt Shopify pipeline, which extracts and loads Shopify data into your data warehouse.

https://github.com/dlt-hub/dbt-dlt-shopify/

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability
- [] The package uses the cross-database macros built into dbt Core where available, such as `{{ dbt.except() }}` and `{{ dbt.type_string() }}`. (It does not use macros)

### Versioning

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -37,7 +37,7 @@
     ],
     "dlt-hub": [
         "dbt_ga4",
-        "dbt-dlt-shopify"
+        "dlt-dbt-shopify"
     ],
     "EqualExperts": [
         "dbt-unit-testing"
--- a/hub.json
+++ b/hub.json
@@ -36,7 +36,8 @@
         "dbt-dataquality"
     ],
     "dlt-hub": [
-        "dbt_ga4"
+        "dbt_ga4",
+        "dbt-dlt-shopify"
     ],
     "EqualExperts": [
         "dbt-unit-testing"


dbeatty10 (approved):
@dat-a-man Looks good 👍 

Non-blocking, but did you intend to include the `logs` and `target` folders? These are typically excluded from GitHub via a [`.gitignore`](https://github.com/dbt-labs/dbt-utils/blob/main/.gitignore) file.

One other thing to note: my favorite dbt command is [`dbt build`](https://docs.getdbt.com/reference/commands/build). It does the equivalent of `dbt run` and `dbt test` in one handy command ([plus](https://docs.getdbt.com/reference/commands/build) seeds and snapshots).

dbeatty10:
@dat-a-man something else to note:
- this package won't show up on the dbt Package Hub until you have at least one semantically versioned tag like [these](https://github.com/TasmanAnalytics/tasman-identity-graph/tags).

dat-a-man:
Thank you for the approval and merging @dbeatty10, 
I've included your comments. 
- Logs and Target were not intended and so they were removed. 
- Documentation updated for dbt build
- Tag v0.1.1 was added

PR #342, opened 2024-11-11
Add dbt-set-similarity

Matts52:
## Description 

This package provides common methods for measuring the similarity of two unordered distinct item sets represented as array/variant type data.

Currently, the following similarity measures are supported:
- Dice-Sorenson Coefficient
- Jaccard Index/Coefficient
- Overlap Coefficient

Sequence/vector similarity is currently considered out of scope for this package but may be considered in future either in this package or as part of a separate package.

Link to your package's repository: https://github.com/Matts52/dbt-set-similarity

## Checklist

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning


dbeatty10 (approved):
Congrats on your 2nd dbt package @Matts52 

Cool package of set similarity metrics 🤩

Note: Currently, the only GitHub tag for this repo is [`release`](https://github.com/Matts52/dbt-set-similarity/tags). You'll need to add a tag with a semantic version similar to like you have in [`dbt-ml-inline-preprocessing`](https://github.com/Matts52/dbt-ml-inline-preprocessing/tags). Your new package won't show up in the [dbt Package Hub](https://hub.getdbt.com/Matts52/) until about an hour or so after you add that tag.

dbeatty10:
@Matts52 Side note: I didn't check one way or the other, but if you implement a macro for the [Tversky index](https://en.wikipedia.org/wiki/Tversky_index), then you might be able to re-use it in the implementations for Jaccard and Dice since it is a generalization of each.

Matts52:
Nice idea for abstraction @dbeatty10 ! I think computationally it may be a bit more burdensome to call that Tversky macro from within the Jaccard/Dice macros vs the single piece denominator found in the explicit Jaccard and Dice formulas. I'll do a little testing and see if I can get clever. 

Nonetheless will implement the Tversky in any case and fix up that release issue before merging

Matts52:
Alright those changes have been made, should be good to merge!

dbeatty10:
Merged, deployed, and available on the dbt Package Hub:

https://hub.getdbt.com/Matts52/

image

PR #369, opened 2025-07-04
Add error protection and handling so a bad project doesn't crash everything

joellabes:
### Add error handling
Added error handling with Claude. Prompt was to protect against failures from external user input such as a git repo not existing or a dbt_project.yml not being present. 

Result of this is that a single invalid package will not be added to the hub, but it won't block everything else from being accepted. 

### Run hubcap when a new package is added
Added a GHA to run hubcap in test mode, then check the logs for error messages. This is useful to do things like:
- validate that the package exists
- validate all package versions contain a valid dbt_project.yml file

Amusingly, while testing it I found that one of the currently open PRs _had_ a failure which it captured! 

### Review tip

The diff looks scarier than it is - I recommend using the Hide Whitespace setting, as everything has been indented inside of try blocks. 
image


joellabes:
Example output: 
image

hubcap/package.py
```diff
@@ -3,51 +3,160 @@
...
+from hubcap.git_helper import clone_repo, run_cmd, GitOperationError
+
+
+class PackageError(Exception):
+    """Custom exception for package operation failures"""
+
+    pass
 
 
 def clone_package_repos(package_maintainer_index, path):
```
b-per, line 22: Not a huge deal, but could add type hint of the return value as `List[str]` as we now return `failed_repos`

PR #417, opened 2026-04-27
Add 'dbt_stackadapt' to the hub.json file

fivetran-joemarkiewicz:
## Description 

Drafting this PR ahead of dbt package completion. Will move out of draft once ready.

This will be a new Fivetran dbt package for the StackAdapt connector. This package will enabled users to better understand the performance of their ads across varying grains and create modeled tables that leverage StackAdapt data. This package will create enriched models with metrics focused on advertiser, campaign, campaign_group, and ad level reports.

Link to your package's repository: https://github.com/fivetran/dbt_stackadapt

## Checklist

### Real world usage

- [ ] (Required): I have been using this package in production and am satisfied with its behavior.

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning
- [x] (Required): The package's git tags validates against the regex defined in `hubcap/version.py` ([examples](https://regex101.com/r/OHA92c)). 

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -353,7 +353,8 @@
         "dbt_ga4_export",
         "dbt_unified_rag",
         "dbt_amazon_selling_partner",
-        "dbt_dynamics_365_crm"
+        "dbt_dynamics_365_crm",
+        "dbt_stackadapt"
     ],
     "fullstorydev": [
         "dbt_fullstory"


fivetran-joemarkiewicz:
@dbeatty10 @joellabes let me know if you have any questions with this PR. Thanks!

joellabes (approved):

PR #427, opened 2026-06-07
Add Cortex Agent Materialization Package `dbt-cortex-agent`

Matts52:
## Description 

Similar to the existing `dbt-semantic-view` package that materializes Snowflake Semantic View packages via dbt, this package materializes Snowflake Cortex Agent packages via dbt. 

The main benefits are:
- version controlling your cortex agents
- having cortex agents in your node lineage
- defining cortex agents within your pipeline

I've tested out that this pckage was able to successfulyl materialize a cortex agent on a Snowflake project at the current sql specification for cortex agents. 

A note is that since snowflake cortex agents only exist on Snowflake, this package only is relevant to the snowflake dbt adapter.

Link to your package's repository: https://github.com/Matts52/dbt-cortex-agent

## Checklist

### Real world usage

- [X] (Required): I have been using this package in production and am satisfied with its behavior.

### First run experience

### Customisability

#### Dependencies on dbt Core
### Interoperability

### Versioning
- [X] (Required): The package's git tags validates against the regex defined in `hubcap/version.py` ([examples](https://regex101.com/r/OHA92c)). 

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -90,7 +90,8 @@
         "dbt-ml-inline-preprocessing",
         "dbt-set-similarity",
         "dbt-stat-test",
-        "dbt-ml-eval"
+        "dbt-ml-eval",
+        "dbt-cortex-agent"
     ],
     "metaplane" : [
         "dbt-expectations"


Matts52:
Hey @joellabes @dbeatty10 , I ended up implementing this package yesterday over at the project at Sigma soundly. Would definitely prefer to import from the package hub, package should be extensible to all Snowflake shops, adding a new object primitive to build with

mattsenicksigma:
Hey @joellabes @dbeatty10 , ended up sharing this package on a webinar with about 100 customer users. Lmk if there is some way we can get this thru to be better versioned via the package hub. (just hate giving people a git commit id to install)

joellabes (approved):
Sorry @Matts52! Thanks for the bump

joellabes:
Trying to force the CI to run...

mattsenicksigma:
sounds good @joellabes , lmk when merged and I can update the README to the package hub install as opposed to the git repo install

joellabes:
Should deploy at the top of the next hour!

PR #428, opened 2026-06-14
feat: add hasbi93/hsbi-dbt-artifacts package

hasbi93:
Description

This package is designed to help dbt practitioners parse, track, and monitor dbt artifacts (such as manifest.json, run_results.json, and sources.json) directly within their target data warehouses. It enables teams to observe execution performance, build histories, and model configurations over time.

Link to your package's repository: https://github.com/hasbi93/hsbi-dbt-artifacts

Checklist

Real world usage
[x] (Required): I have been using this package in production and am satisfied with its behavior.

First run experience
[x] (Required): The package includes a [licence file detectable by GitHub](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository), such as the Apache 2.0 or MIT licence.
[x] The package includes a README which explains how to get started with the package and customise its behaviour.
[x] The README indicates which data warehouses/platforms are expected to work with this package.

Customisability
[x] The package uses ref or source, instead of hard-coding table references.

Packages for data transformation (delete if not relevant):
[x] Provide a mechanism (such as variables) to customise the location of source tables.
[x] Do not assume database/schema names in sources.

Dependencies
Dependencies on dbt Core
[x] The package has set a supported require-dbt-version range in dbt_project.yml.

Dependencies on other packages defined in packages.yml:
[x] Dependencies are imported from the dbt Package Hub when available, as opposed to a git installation.
[x] Dependencies contain the widest possible range of supported versions, to minimise issues in dependency resolution.
[x] In particular, dependencies are not pinned to a patch version unless there is a known incompatibility.

Interoperability
[x] The package does not override dbt Core behaviour in such a way as to impact other dbt resources (models, tests, etc) not provided by the package.
[x] The package uses the cross-database macros built into dbt Core where available, such as {{ dbt.except() }} and {{ dbt.type_string() }}.
[x] The package disambiguates its resource names to avoid clashes with nodes that are likely to already exist in a project.

Versioning
[x] (Required): The package's git tags validates against the regex defined in hubcap/version.py.
[x] The package's version follows the guidance of Semantic Versioning 2.0.0.

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -362,6 +362,9 @@
     "gauthierpiarrette": [
         "dbt-timefence"
     ],
+    "hasbi93": [
+        "hsbi-dbt-artifacts"
+    ],
     "get-select": [
         "dbt-query-tags",
         "dbt-snowflake-monitoring",


dbeatty10:
@hasbi93 To the best that I can tell, the only difference between [`hasbi93/hsbi-dbt-artifacts`](https://github.com/hasbi93/hsbi-dbt-artifacts) and [`brooklyn-data/dbt_artifacts`](https://github.com/brooklyn-data/dbt_artifacts) is this:

```diff
- , cast(null as {{ type_json() }}) as all_results
+, cast(null as VARCHAR(MAX)) as all_results
```

Our intent for the package hub is for each package to be unique rather than forks with small changes.

So I'm going to close this in favor of the following suggestion instead:

1. [Open a bug report with `brooklyn-data/dbt_artifacts`](https://github.com/brooklyn-data/dbt_artifacts/issues/new?template=bug_report.md) explaining the issue you are running into to see if it can be addressed.

2. Optional: In the meantime, you can use the `brooklyn-data/dbt_artifacts` package and override the relevant model definitions like [this](https://docs.getdbt.com/reference/resource-configs/enabled?version=2.0&name=v2#disable-a-model-in-a-package-in-order-to-use-your-own-version-of-the-model).

PR #435, opened 2026-07-21
Add Elsayed91/dbt-subzero

elsayed91:
Adds [dbt-subzero](https://github.com/Elsayed91/dbt-subzero): data quality tests for Apache Iceberg tables answered from table metadata (manifest statistics, snapshots) instead of scanning data. Supports dbt-trino and dbt-spark. First release v0.1.0 is published.

---

## Checklist

### Real world usage

- [ ] (Required): I have been using this package in production and am satisfied with its behavior.

### First run experience

### Customisability

#### Dependencies on dbt Core
Zero dependencies 

### Interoperability

### Versioning
- [x] (Required): The package's git tags validates against the regex defined in `hubcap/version.py` ([examples](https://regex101.com/r/OHA92c)). 

### DIFF
--- a/hub.json
+++ b/hub.json
@@ -11,6 +11,9 @@
     "AltimateAI": [
         "altimate-dbt-snowflake-query-tags"
     ],
+    "Elsayed91": [
+        "dbt-subzero"
+    ],
     "M-KOPA": [
         "insert_by_timeperiod"
     ],


dbeatty10:
@elsayed91 I've added the [checklist](https://github.com/dbt-labs/hubcap/blob/a5cf0e73cf7acd611894b03f40ab09d33ab737fa/.github/pull_request_template.md#L7) to the PR description.

Could you fill it out to help us to review your package submission?

elsayed91:
hi @dbeatty10 

It has not been running in production long enough but I am satisfied with the behavior, tbh the premise is rather simple it is just calculating some data quality tests by querying iceberg table metadata rather that doing full scans. there is no state, and no dependency beyond metadata tables that are exposed by the engines. 

We can let it simmer for some time however if you feel like that is the right call

PR #448, opened 2026-09-02
Add yu-iskw/dbt-bigquery-federation package

yu-iskw:
## Description

`dbt_bigquery_federation` makes BigQuery `EXTERNAL_QUERY` practical from dbt. It plans type conversions from Git-reviewed schema pins and renders federated table expressions for Cloud SQL PostgreSQL, AlloyDB PostgreSQL, and Spanner GoogleSQL.

https://docs.cloud.google.com/bigquery/docs/federated-queries-intro

Hub identity after indexing:
- GitHub repo: `yu-iskw/dbt-bigquery-federation`
- Hub package name: `yu-iskw/dbt_bigquery_federation`

A semver git tag (for example `v0.1.0`) still needs to be created so hubcap can publish the first version after merge.

Link to your package's repository: https://github.com/yu-iskw/dbt-bigquery-federation

## Checklist

### Real world usage

- [x] (Required): I have been using this package in production and am satisfied with its behavior.

### First run experience

### Customisability

This is a macro package, not a data-transformation package. Call sites pass `connection` / `schema` / `table`, and pins live in `vars.dbt_bigquery_federation`.

#### Dependencies on dbt Core

No root `packages.yml` dependencies.

### Interoperability

### Versioning
- [ ] (Required): The package's git tags validates against the regex defined in `hubcap/version.py` ([examples](https://regex101.com/r/OHA92c)).


dbeatty10 (approved):
🚀

PR #449, opened 2026-09-02
Mirror release tarballs to S3 and record downloads.hub in specs

tauhid621:
## Why
GitHub tags aren't immutable — a maintainer can move a tag to point at a different commit after we've already published a version pointing at its tarball. Since the tarball URL resolves the tag's contents at download time, that move silently changes what an already-"released" version actually installs. Mirroring the tarball at the moment we first see the tag pins those exact bytes, so a version stays what it was when it was published, even if the upstream tag moves later.

## What changes
- Every new package version now has its release tarball mirrored to dbt Labs' own storage the moment it's first published, capturing exactly what shipped at that time.
- The version spec gains a second download location (`downloads.hub`) pointing at that mirrored copy, alongside the existing GitHub link (`downloads.tarball`) — same file, verified by matching checksums.
- If mirroring isn't configured, or a mirror upload fails for any reason, publishing continues exactly as it did before — the extra field is simply absent. No package can be blocked from going out because of mirroring trouble.
- Re-running hubcap won't re-mirror a tarball that's already been copied.
- Mirroring credentials are short-lived and obtained per-run rather than stored as a long-lived secret; PR runs from forks skip mirroring entirely since they don't get those credentials — same as today.

## Testing
- Unit tests cover the mirroring logic and its failure/fallback paths (178 tests passing).
- Ran the real pipeline end to end against a staging environment: published a real package version, and independently verified the resulting hub download URL returns the tarball (HTTP 200, correct content type) with content that hashes to the exact checksum recorded in the spec.


dbeatty10 (approved):
Looks amazing! 🚀

Not blocking for this PR:
- we should do some manual testing to make sure the `DBT_ENGINE_PACKAGE_HUB_URL` environment variable works for self-hosting an internal version of the package hub with and without S3 bucket included in `HUBCAP_CONFIG` / `config.json`.

PR #452, opened 2026-09-08
Add script and workflow to backfill downloads.hub on existing specs

tauhid621:
Backfills `downloads.hub` on specs published before the S3 mirroring in #449, mirroring the tarball to S3 and updating just that field. Runs via manual `workflow_dispatch` (dry-run by default), processes a bounded batch, and opens one follow-up PR per run with the updated specs.


dbeatty10 (approved):
🚀

